PR #10, opened 2017-07-27
Update VAL/VEDA Course ID sharing

yro:
### Description
Updates VAL/VEDA course_id handling

### Testing
- [x] Unit tests

### Reviewers
- [ ] Code review: @muzaffaryousaf @Qubad786 

FYI: @edx/educator-neem

### Post-review
- [x] Rebase and squash commits

### DIFF
--- a/control/veda_val.py
+++ b/control/veda_val.py
@@ -356,9 +356,12 @@ class VALAPICall():
         VAL will not allow duped studio urls to be sent,
         so we must scrub the data
         """
-        for c in val_api_return['courses']:
-            if c in self.val_data['courses']:
-                self.val_data['courses'].remove(c)
+        for course in val_api_return['courses']:
+            for key, entry in course.items():
+                for crs in self.val_data['courses']:
+                    for ky, ey in crs.items():
+                        if ky == key:
+                            self.val_data['courses'].remove(crs)
 
         self.profile_determiner(val_api_return=val_api_return)
         self.val_data['status'] = self.val_status
--- a/control/veda_val.py
+++ b/control/veda_val.py
@@ -357,11 +357,10 @@ class VALAPICall():
         so we must scrub the data
         """
         for course in val_api_return['courses']:
-            for key, entry in course.items():
-                for crs in self.val_data['courses']:
-                    for ky, ey in crs.items():
-                        if ky == key:
-                            self.val_data['courses'].remove(crs)
+            for course_id in course.keys():
+                for course_entry in self.val_data['courses']:
+                    if course_id in course_entry:
+                        self.val_data['courses'].remove(course_entry)
 
         self.profile_determiner(val_api_return=val_api_return)
         self.val_data['status'] = self.val_status


control/veda_val.py
```diff
@@ -356,9 +356,12 @@ def send_200(self, val_api_return):
         VAL will not allow duped studio urls to be sent,
         so we must scrub the data
         """
-        for c in val_api_return['courses']:
-            if c in self.val_data['courses']:
-                self.val_data['courses'].remove(c)
+        for course in val_api_return['courses']:
+            for key, entry in course.items():
```
iloveagent57: if you're never going to use `entry`, just iterate over `course.keys()`.

control/veda_val.py
```diff
@@ -356,9 +356,12 @@ def send_200(self, val_api_return):
...
+                for crs in self.val_data['courses']:
```
iloveagent57: What, conceptually, are the things in `self.val_data['courses']`? Are they existing courses?  If so, maybe a better  variable name for `crs` is `existing_course`, or something like that.

control/veda_val.py
```diff
@@ -356,9 +356,12 @@ def send_200(self, val_api_return):
...
+                    for ky, ey in crs.items():
```
iloveagent57: You don't use `ey` anywhere, so use `keys()` instead of `items()`.  Also, `ky` could be replaced by something more descriptive.

control/veda_val.py
```diff
@@ -356,9 +356,12 @@ def send_200(self, val_api_return):
...
+                        if ky == key:
+                            self.val_data['courses'].remove(crs)
```
iloveagent57: You can `break` after this line - you don't have to check any more keys in the course once you've removed the course.

yro:
@iloveagent57 Thank you so much for the input!

control/veda_val.py
yro: @iloveagent57 That's not correct. There's a possibility that a course entry can get duplicated in the retrieved data from the database, so the entire array has to be looped over.

Qubad786: you may not want to remove while iterating through the same `list`.

control/veda_val.py
```diff
@@ -356,9 +356,12 @@ def send_200(self, val_api_return):
...
         so we must scrub the data
         """
-        for c in val_api_return['courses']:
-            if c in self.val_data['courses']:
-                self.val_data['courses'].remove(c)
+        for course in val_api_return['courses']:
+            for key, entry in course.items():
+                for crs in self.val_data['courses']:
+                    for ky, ey in crs.items():
+                        if ky == key:
```
Qubad786: you may not want to remove while iterating through the same `list`.

control/veda_val.py
```diff
@@ -356,9 +356,11 @@ def send_200(self, val_api_return):
         VAL will not allow duped studio urls to be sent,
         so we must scrub the data
         """
-        for c in val_api_return['courses']:
-            if c in self.val_data['courses']:
-                self.val_data['courses'].remove(c)
+        for course in val_api_return['courses']:
+            for course_id in course.keys():
+                for course_entry in self.val_data['courses']:
```
Qubad786: This is not good, we shouldn't iterate and remove on the same list, it would give unexpected results.

control/veda_val.py
```diff
@@ -356,9 +356,11 @@ def send_200(self, val_api_return):
         VAL will not allow duped studio urls to be sent,
         so we must scrub the data
         """
-        for c in val_api_return['courses']:
-            if c in self.val_data['courses']:
-                self.val_data['courses'].remove(c)
+        for course in val_api_return['courses']:
+            for course_id in course.keys():
```
Qubad786: its not necessary to loop on course.keys(), see the source: https://github.com/edx/edx-val/blob/141f888462f7a93c01a69ab1340f7e4481e0c085/edxval/serializers.py#L95

yro:
@Qubad786 I am unassigning myself from this ticket. It seems like you have a good fix.

PR #14, opened 2017-08-01
Save organisation transcript preferences - EDUCATOR-1048

mushtaqak:
The purpose of this PR is to save organisation transcript preferences and show in admin side.

[EDUCATOR-1048](https://openedx.atlassian.net/browse/EDUCATOR-1048)


VEDA_OS01/models.py
```diff
@@ -428,3 +428,31 @@ def __unicode__(self):
...
+        (THREE_PLAY, THREE_PLAY),
+        (CIELO24, CIELO24),
+    )
+
+
+class TranscriptionProvider(models.Model):
+    """
+    Model to contain third party transcription service provider preferances.
+    """
+    org = models.CharField('Organization', max_length=50, unique=True)
```
muzaffaryousaf: can we add help text to tell others that these are organizations from studio / platform, not the institutes in VEDA.

VEDA_OS01/models.py
```diff
@@ -428,3 +428,31 @@ def __unicode__(self):
...
+    provider = models.CharField('Transcription provider', max_length=50, choices=TranscriptionProviderType.CHOICES)
+    api_key = models.CharField('API key', max_length=255)
+    api_secret = models.CharField('API secret', max_length=255, null=True, blank=True)
+    created = models.DateTimeField(auto_now_add=True, null=True)
+    modified = models.DateTimeField(auto_now=True, null=True)
```
muzaffaryousaf: can we use timestamped model which includes these fields by default ?

mushtaqak: timestamped is not available in this repo. we can either add in or just go with these 2 fields for now.

muzaffaryousaf: Let's add Django model utils to use timestamped model.

VEDA_OS01/models.py
```diff
@@ -428,3 +428,31 @@ def __unicode__(self):
...
+        (CIELO24, CIELO24),
+    )
+
+
+class TranscriptionProvider(models.Model):
+    """
+    Model to contain third party transcription service provider preferances.
+    """
+    org = models.CharField('Organization', max_length=50, unique=True)
+    provider = models.CharField('Transcription provider', max_length=50, choices=TranscriptionProviderType.CHOICES)
```
muzaffaryousaf: @mushtaqak I think we need to make org & provider unique in a combination. So that orgs can have both credentials to use them in different courses.

VEDA_OS01/models.py
```diff
@@ -428,3 +428,31 @@ def __unicode__(self):
...
+    api_key = models.CharField('API key', max_length=255)
+    api_secret = models.CharField('API secret', max_length=255, null=True, blank=True)
+    created = models.DateTimeField(auto_now_add=True, null=True)
+    modified = models.DateTimeField(auto_now=True, null=True)
+
+    def __unicode__(self):
+        return self.org
```
muzaffaryousaf: I would add org & provider as unicode. Also, can we make this real unicode.

VEDA_OS01/models.py
```diff
@@ -428,3 +428,31 @@ def __unicode__(self):
...
+
+    THREE_PLAY = '3PlayMedia'
+    CIELO24 = 'Cielo24'
+    CHOICES = (
+        (THREE_PLAY, THREE_PLAY),
+        (CIELO24, CIELO24),
+    )
+
+
+class TranscriptionProvider(models.Model):
```
muzaffaryousaf: Can we name it `TranscriptionPreferences` ?

muzaffaryousaf:
@yro I can't see migration in edx-video-pipeline, is it by intention ??? Also, how they work on production ?

VEDA_OS01/models.py
mushtaqak: Added

VEDA/settings.py
```diff
@@ -137,6 +137,7 @@
     'rest_framework.authtoken',
     'oauth2_provider',
     'rest_framework',
+    'django_filters',
```
mushtaqak: **Note to reviewers:**  This is added to resolve `TemplateNotExist` error when accessing rest api views.

See https://github.com/edx/credentials/pull/284/files and carltongibson/django-filter#562

yro:
@mushtaqak We don't have a migration folder yet. I'll do the migrations (via manage.py) for this on deploy and check the folder back into the repo for future use. Good catch!

muzaffaryousaf:
@yro Do you think we should merge the changes related to models in master branch without migrations. I'm not sure just because of migrations on prod and how they'll work.

yro:
@muzaffaryousaf yes you should merge without migrations _just this one time_. After I push this PR to prod, I'll pull the migrations back into the repo, and future changes to models/migrations can be made in the repo.

VEDA_OS01/models.py
```diff
@@ -463,3 +480,30 @@ class Meta:
...
+    lang_code = models.CharField('Language code', max_length=3)
+    status = models.CharField(
+        'Transcript status',
+        max_length=50,
+        choices=TranscriptStatus.CHOICES,
+        default=TranscriptStatus.PENDING
+    )
+
+    class Meta:
+        unique_together = ('video', 'provider', 'lang_code')
```
Qubad786: TODO: we need to remove this constraint, this will raise integrity in case of VEDA heal (i.e. transcription re-process for same lang)

PR #18, opened 2017-08-02
Add transcript preferences in Video

Qubad786:
1. Add transcript preferences in Video
2. Expect the transcript preferences in file discovery phase
3. Pass these preferences from file discovery to ingestion phase for cataloging

@muzaffaryousaf

### DIFF
--- a/VEDA_OS01/models.py
+++ b/VEDA_OS01/models.py
@@ -1,6 +1,7 @@
 """
 Models for Video Pipeline
 """
+import json
 import uuid
 from django.db import models
 from model_utils.models import TimeStampedModel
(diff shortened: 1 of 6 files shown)


muzaffaryousaf:
@Qubad786 Will take a look in morning.

VEDA_OS01/models.py
```diff
@@ -40,6 +41,111 @@ class TranscriptStatus(object):
...
+        Returns:
+            A list if validation is successful
+
+        Raises:
+            ValidationError
+        """
+        if all(isinstance(item, basestring) for item in value) is False:
+            raise ValueError(u'list must only contain strings.')
+
+        return value
```
muzaffaryousaf: I'm assuming this code is already being reviewed and tested. Let me know if that's not the case.

VEDA_OS01/models.py
```diff
@@ -40,6 +41,111 @@ class TranscriptStatus(object):
...
+    RUSH_SERVICE = 'rush_service'
+    SAME_DAY_SERVICE = 'same_day_service'
+
+    CHOICES = (
+        (EXTENDED_SERVICE, '10-Day/Extended'),
+        (DEFAULT, '4-Day/Default'),
+        (EXPEDITED_SERVICE, '2-Day/Expedited'),
+        (RUSH_SERVICE, '24 hour/Rush'),
+        (SAME_DAY_SERVICE, 'Same Day'),
+    )
```
muzaffaryousaf: @Qubad786 Since you've added these classes, i think we should remove the existing choices tuples/ variables from course model. Let's keep it simple and clean.

VEDA_OS01/models.py
```diff
@@ -317,6 +423,33 @@ class Video (models.Model):
...
+        choices=Cielo24Turnaround.CHOICES,
+        null=True
+    )
+    cielo24_fidelity = models.CharField(
+        'Cielo24 Fidelity',
+        max_length=20,
+        choices=Cielo24Fidelity.CHOICES,
+        null=True
+    )
+    preferred_languages = ListField(default=[])
```
muzaffaryousaf: Purpose of `c24_hours` in previous model ?

control/veda_file_discovery.py
```diff
@@ -252,15 +256,39 @@ def studio_s3_validate(self, meta, key):
...
+            transcript_preferences = json.loads(transcript_preferences)
+            TranscriptionPreferences.objects.get(
+                org=transcript_preferences.get('org'),
+                provider=transcript_preferences.get('provider')
+            )
+            process_transcription = True
+        except (TypeError, ValueError, TranscriptionPreferences.DoesNotExist):
+            # when these preferences are not set, set to some invalid data or these don't
+            # have associated 3rd party transcription provider's API secrets.
+            process_transcription = False
```
muzaffaryousaf: @Qubad786 Can we add logging here, to see what's causing the exception.

control/veda_file_ingest.py
```diff
@@ -342,6 +349,15 @@ def database_record(self):
...
             return None
 
+        # Update transcription preferences for the Video
+        if self.video_proto.process_transcription:
+            v1.process_transcription = self.video_proto.process_transcription
+            v1.provider = self.video_proto.provider
+            v1.three_play_turnaround = self.video_proto.three_play_turnaround
+            v1.cielo24_turnaround = self.video_proto.cielo24_turnaround
+            v1.cielo24_fidelity = self.video_proto.cielo24_fidelity
+            v1.preferred_languages = self.video_proto.preferred_languages
```
muzaffaryousaf: can we add loggs for except block where we're doing v1.save at line 380.

VEDA_OS01/models.py
Qubad786: that's turnaround override which automatically controls the `PRIORITY`

VEDA_OS01/models.py
Qubad786: yes, its already being used and tested for video-images

VEDA_OS01/models.py
Qubad786: I think we're going to deprecate them at once? if so, why would we need to clean?

muzaffaryousaf: If classes are same then why not remove them now, later part would be only migration and removing fields.

control/veda_file_discovery.py
muzaffaryousaf: Ignore this comment as per discussion. Reason: the logs would be noisy.

VEDA_OS01/models.py
muzaffaryousaf: Do we need cater in our design ?

muzaffaryousaf:
@Qubad786 We need to rebase this before the final look.

VEDA_OS01/models.py
```diff
@@ -255,68 +426,48 @@ class Video (models.Model):
...
-        ('Complete', "Complete")
-
-    )
     video_trans_status = models.CharField(
         'Transcode Status',
         max_length=100,
-        choices=TRANS_STATUS_OPTIONS,
-        default=SI
+        choices=VideoStatus.CHOICES,
+        default=VideoStatus.SI
```
muzaffaryousaf: @Qubad786 Are you sure to change this, this will require an migration too.

VEDA_OS01/models.py
```diff
@@ -255,68 +426,48 @@ class Video (models.Model):
...
+    )
+    three_play_turnaround = models.CharField(
+        '3PlayMedia Turnaround',
+        max_length=20,
+        choices=ThreePlayTurnaround.CHOICES,
+        null=True,
+        blank=True,
+    )
+    cielo24_turnaround = models.CharField(
+        'Cielo24 Turnaround', max_length=20,
```
muzaffaryousaf: hope you've verified the max_length for all these fields, i would prefer to have a considerable cushion in max_length for these fields.

control/veda_file_discovery.py
```diff
@@ -252,24 +256,45 @@ def studio_s3_validate(self, meta, key):
...
-        """
-        Trigger Ingest Process
-        """
-        V = VideoProto(
+        # Make decision if this video needs the transcription as well.
+        try:
+            transcript_preferences = json.loads(transcript_preferences)
+            TranscriptPreferences.objects.get(
+                # TODO: Once ammar is done with cielo24.
+                #org=extract_course_org(course_url),
```
muzaffaryousaf: can we leave org as it for now along with this comment and TODO

control/veda_file_ingest.py
```diff
@@ -377,6 +397,12 @@ def database_record(self):
             v1.client_title = final_string
             v1.save()
 
+        except Exception:
+            # Log the exception and raise.
+            LOGGER.exception('[Ingest] Cataloging of a video=%s failed.', self.video_proto.veda_id)
```
muzaffaryousaf: can we prefix these logs with something VIDEO PIPELINE ?

muzaffaryousaf:
@Qubad786 When you make suggested changes, make sure that upstream branch is up-to-date with master.

VEDA_OS01/models.py
muzaffaryousaf: Since we're adding new choices so we need it anyway.

VEDA_OS01/models.py
```diff
@@ -255,68 +426,48 @@ class Video (models.Model):
...
+        blank=True,
+    )
+    cielo24_fidelity = models.CharField(
+        'Cielo24 Fidelity',
+        max_length=20,
+        choices=Cielo24Fidelity.CHOICES,
+        null=True,
+        blank=True,
+    )
+    preferred_languages = ListField(default=[])
```
muzaffaryousaf: If this is optional then we've to specify blank=True for this field.

control/veda_file_discovery.py
```diff
@@ -252,24 +256,45 @@ def studio_s3_validate(self, meta, key):
...
+                provider=transcript_preferences.get('provider')
+            )
+            process_transcription = True
+        except (TypeError, ValueError, TranscriptPreferences.DoesNotExist):
```
muzaffaryousaf: can we add logs for ValueError only, in case when there's invalid data.

control/veda_file_discovery.py
```diff
@@ -252,24 +256,45 @@ def studio_s3_validate(self, meta, key):
...
-        Trigger Ingest Process
-        """
-        V = VideoProto(
+        # Make decision if this video needs the transcription as well.
+        try:
+            transcript_preferences = json.loads(transcript_preferences)
+            TranscriptPreferences.objects.get(
+                # TODO: Once ammar is done with cielo24.
+                #org=extract_course_org(course_url),
+                provider=transcript_preferences.get('provider')
```
muzaffaryousaf: We should't hit the DB if provider has no value.

VEDA_OS01/models.py
Qubad786: I've seen not a single choice crossing 20 chars length.

control/veda_file_discovery.py
Qubad786: okay

Qubad786:
@muzaffaryousaf addressed everything.

VEDA_OS01/models.py
muzaffaryousaf: @Qubad786  I'm not seeing blank=True for preferred_languages field. this will break the migrations on production for existing records.

muzaffaryousaf: Let's add null & blank both True.

control/veda_file_ingest.py
muzaffaryousaf: Addressed.

muzaffaryousaf:
LGTM 👍

PR #25, opened 2017-08-16
green build

muhammad-ammar:
@muzaffaryousaf @yro These changes will make the build green. Please have a look.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,15 @@
 PACKAGES = VEDA VEDA_OS01 control frontend youtube_callback scripts
 
+requirements:
+	pip install -r requirements.txt
+	pip install -r test_requirements.txt
+
 validate: test ## Run tests and quality checks
 
 test: clean
-	 nosetests --with-coverage --cover-inclusive --cover-branches \
-		--cover-html --cover-html-dir=build/coverage/html/ \
-		--cover-xml --cover-xml-file=build/coverage/coverage.xml --verbosity=2 \
-		$(foreach package,$(PACKAGES),--cover-package=$(package)) \
-		$(PACKAGES)
+	coverage run -m pytest --durations=10
+	coverage combine
+	coverage report
 
 clean:
 	coverage erase
--- a/control/tests/test_heal.py
+++ b/control/tests/test_heal.py
@@ -1,36 +1,90 @@
-
-import os
-import sys
-import unittest
-
 """
 Test heal processor
-
 """
+import datetime
+import json
+import os
+import sys
+from django.test import TestCase
+from datetime import timedelta
+
+import responses
+from django.utils.timezone import utc
+from mock import PropertyMock, patch
+
+from control.veda_heal import VedaHeal
+from VEDA_OS01.models import URL, Course, Destination, Encode, Video
+from VEDA_OS01.utils import build_url, get_config
+
 sys.path.append(os.path.dirname(os.path.dirname(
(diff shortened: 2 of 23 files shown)


Makefile
```diff
@@ -1,13 +1,15 @@
...
 
 test: clean
-	 nosetests --with-coverage --cover-inclusive --cover-branches \
-		--cover-html --cover-html-dir=build/coverage/html/ \
-		--cover-xml --cover-xml-file=build/coverage/coverage.xml --verbosity=2 \
-		$(foreach package,$(PACKAGES),--cover-package=$(package)) \
-		$(PACKAGES)
+	DJANGO_SETTINGS_MODULE=VEDA.test_settings coverage run -m pytest --durations=10
+	coverage combine
+	coverage report --fail-under=49
```
muhammad-ammar: Note to Reviewers: Remove `--fail-under=49` once we have `codecov` integration

muhammad-ammar:
CC: @Qubad786 @mushtaqak

muzaffaryousaf:
@muhammad-ammar can we make this PR against master ? I'm not sure when we'll merge the feature branch into master but it would be nice if we merge this ASAP in master to avoid any failing changes in master.

muzaffaryousaf:
@muhammad-ammar fyi ^^ ???

muhammad-ammar:
@muzaffaryousaf I would prefer to merge it in `ammar/update-cielo24-integration`. There are some changes specific to `ammar/update-cielo24-integration` and I do not want to make those changes again.

Makefile
muzaffaryousaf: codecov was working in case of success. It should work here too. You can double check travis file.

control/tests/test_heal.py
```diff
@@ -1,36 +1,104 @@
...
+    'val_username': 'val_username',
+    'val_password': 'val_password',
+    'val_client_id': 'val_client_id',
+    'val_secret_key': 'val_secret_key',
+    'val_profile_dict': {
+        'desktop_mp4': ['desktop_mp4'],
+        'hls': ['hls'],
+        'mobile_low': ['mobile_low']
+    }
+}
```
muhammad-ammar: This would be great but I am not sure what is the best strategy to handle this? I can think of two ways:
1. Create a new config file(`test_instance_config.yaml`) for tests and then update all the occurrences of `instance_config.yaml` and based on a `settings`, decide which config file we need to load.
2. Update the `instance_config.yaml` with sane default values. Note: My understanding is that for production `instance_config.yaml` is replaced with production settings so there will be no issue in keeping default values.

muhammad-ammar:
@muzaffaryousaf `codecov` job is failing because target coverage is not achieved. Should we change the target coverage in `codecov.yml` or merge the failing PR?

control/tests/test_heal.py
muzaffaryousaf: @muhammad-ammar can we checkin edx-video-worker how we're achieving same thing.

muzaffaryousaf:
> @muzaffaryousaf codecov job is failing because target coverage is not achieved. Should we change the target coverage in codecov.yml or merge the failing PR?

Let's merge it with failing coverage.

muhammad-ammar:
@muzaffaryousaf I have added `test_config.yaml`. Good to merge?

muzaffaryousaf:
👍

PR #34, opened 2017-09-27
Add JWT settings

mushtaqak:
[EDUCATOR-1335](https://openedx.atlassian.net/browse/EDUCATOR-1335)

[Related platform PR](https://github.com/edx/edx-platform/pull/16116)

### DIFF
--- a/VEDA/settings/base.py
+++ b/VEDA/settings/base.py
@@ -110,6 +110,15 @@ DATABASES = {
     }
 }
 
+# NOTE: JWT_SECRET_KEY is intentionally not set here to avoid production releases with a public value.
+# Set a value in a downstream settings file.
+JWT_AUTH = {
+    'JWT_AUDIENCE': 'video-pipeline',
+    'JWT_ISSUER': 'video-pipeline',
+    'JWT_DECODE_HANDLER': 'edx_rest_framework_extensions.utils.jwt_decode_handler',
+    'JWT_VERIFY_AUDIENCE': False,
+}
+
 REST_FRAMEWORK = {
     'PAGE_SIZE': 10,
     'DEFAULT_PARSER_CLASSES': (
@@ -119,6 +128,7 @@ REST_FRAMEWORK = {
         'rest_framework.permissions.IsAuthenticated',
     ),
     'DEFAULT_AUTHENTICATION_CLASSES': (
+        'edx_rest_framework_extensions.authentication.JwtAuthentication',
         'oauth2_provider.ext.rest_framework.OAuth2Authentication',
         'rest_framework.authentication.BasicAuthentication',
         'rest_framework.authentication.SessionAuthentication',
--- a/instance_config.yaml
+++ b/instance_config.yaml
@@ -25,6 +25,14 @@ SECRET_KEY: ""
 # (set to false in prod)
 debug: false
 
+
+# JWT AUTH settings
+JWT_AUTH:
+    JWT_SECRET_KEY':
+    JWT_ISSUER':
+    JWT_AUDIENCE':
+    JWT_VERIFY_AUDIENCE':
+
 # ---
 # AWS Buckets, Prefixes
 # ---
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,11 @@
 django==1.9
 djangorestframework==3.6.4
 django-cors-headers
-django-oauth-toolkit==0.11.0
-django-model-utils==3.0.0
 django-filter==1.0.4
+django-model-utils==3.0.0
+django-oauth-toolkit==0.11.0
+edx-drf-extensions==1.2.3
+edx-opaque-keys==0.4
 newrelic
 uwsgi
 pysftp
@@ -12,7 +14,6 @@ pyyaml
 requests==2.18.1
 celery==3.1.18
 pysrt==1.1.1
-edx-opaque-keys==0.4
 MySQL-python==1.2.5
 gunicorn==0.17.4
 gevent==1.2.2
--- a/VEDA/settings/local.py
+++ b/VEDA/settings/local.py
@@ -17,6 +17,13 @@ DATABASES = {
 }
 # END DATABASE CONFIGURATION
 
+JWT_AUTH.update({
+    'JWT_SECRET_KEY': 'lms-secret',
+    'JWT_ISSUER': 'http://127.0.0.1:8000/oauth2',
+    'JWT_AUDIENCE': 'lms-key',
+    'JWT_VERIFY_AUDIENCE': False
+})
+
 # See if the developer has any local overrides.
 if os.path.isfile(os.path.join(os.path.dirname(os.path.abspath(__file__)), 'private.py')):
     from .private import *  # pylint: disable=import-error, wildcard-import
--- a/VEDA/settings/production.py
+++ b/VEDA/settings/production.py
@@ -10,7 +10,7 @@ ALLOWED_HOSTS = ['*']
 
 # Keep track of the names of settings that represent dicts. Instead of overriding the values in base.py,
 # the values read from disk should UPDATE the pre-configured dicts.
-DICT_UPDATE_KEYS = ('DATABASES', 'SECRET_KEY', 'STATIC_ROOT_PATH')
+DICT_UPDATE_KEYS = ('DATABASES', 'SECRET_KEY', 'STATIC_ROOT_PATH', 'JWT_AUTH')
 
 CONFIG_DATA = get_config()
 # Remove the items that should be used to update dicts, and apply them separately rather


mushtaqak:
@Qubad786 This is ready for your review

VEDA_OS01/transcripts.py
```diff
@@ -360,6 +370,97 @@ def post(self, request, **kwargs):
...
+
+        # TODO: Encode credentials
+        transcript_credentials = {
+            'api_key': api_key,
+            'api_secret': three_play_secret_key
+        }
+
+        try:
+            TranscriptCredentials.objects.update_or_create(
+                org=org, provider=provider, defaults = transcript_credentials
```
Qubad786:
```bash
TranscriptCredentials.objects.update_or_create(
    org=org, provider=provider, defaults={
        'api_key': api_key,
        'api_secret': three_play_secret_key
    }
)
```

VEDA_OS01/transcripts.py
```diff
@@ -360,6 +370,97 @@ def post(self, request, **kwargs):
...
+            )
+        except ValidationError as ex:
```
Qubad786: No need for this.

VEDA_OS01/transcripts.py
```diff
@@ -38,11 +40,19 @@
 CIELO24_TRANSCRIPT_COMPLETED = django.dispatch.Signal(providing_args=['job_id', 'lang_code', 'org', 'video_id'])
 CONFIG = utils.get_config()
 
+
+# Cielo24 API version
+CIELO24_API_VERSION = 1
```
Qubad786: currently, we do not use it in cielo24 callback/client, should this PR also cover that?

VEDA_OS01/transcripts.py
```diff
@@ -360,6 +370,97 @@ def post(self, request, **kwargs):
...
+        if missing:
+            LOGGER.warning(
+                '[Transcript Credentials] Required transcript credentials params are missing. %s', ' and '.join(missing)
+            )
+            return Response(
+                status=status.HTTP_400_BAD_REQUEST,
+                data=dict(message=u'{missing} must be specified.'.format(missing=' and '.join(missing)))
+            )
+
+        org = request.data.get('org')
```
Qubad786: we are unnecessarily defensive here -- we can use `request.data['org']`

VEDA_OS01/transcripts.py
```diff
@@ -360,6 +370,97 @@ def post(self, request, **kwargs):
...
+        """
+        return requests.get(CIELO24_LOGIN_URL, params={
+            'v': CIELO24_API_VERSION,
+            'username': username,
+            'securekey': api_secure_key
+        })
+
+    def post(self, request):
+        """
+        Creates or updates a transcript credential instance with the given information.
```
Qubad786:
```bash
"""
Creates or updates the org-specific transcript credentials with the given information
"""
```

VEDA_OS01/transcripts.py
```diff
@@ -360,6 +370,97 @@ def post(self, request, **kwargs):
...
+        Arguments:
+            request: A WSGI request.
+        """
+        attrs = ('org', 'provider', 'api_key')
+        provider = request.data.get('provider')
+
+        if provider == TranscriptProvider.CIELO24:
+            attrs += ('cielo_username', )
+
+        if provider == TranscriptProvider.THREE_PLAY:
```
Qubad786: provider will be either 3play or Cielo24, we may use `if/else` clause

VEDA_OS01/transcripts.py
```diff
@@ -360,6 +370,97 @@ def post(self, request, **kwargs):
...
+            )
+            return Response(
+                status=status.HTTP_400_BAD_REQUEST,
+                data=dict(message=u'{missing} must be specified.'.format(missing=' and '.join(missing)))
+            )
+
+        org = request.data.get('org')
+        api_key = request.data.get('api_key')
+        cielo_username = request.data.get('cielo_username')
+        three_play_secret_key = request.data.get('three_play_secret_key')
```
Qubad786: shouldn't we also get it conditionally? based on the provider?

VEDA/settings.py
```diff
@@ -110,6 +110,15 @@
 # Python dotted path to the WSGI application used by Django's runserver.
 WSGI_APPLICATION = 'VEDA.wsgi.application'
 
+# See: http://getblimp.github.io/django-rest-framework-jwt/#additional-settings
+JWT_AUTH = {
+    # If the secret is wrong, JWT_VERIFY will raise a jwt.DecodeError telling you as such.
+    # You can still get at the payload by setting the JWT_VERIFY to False
+    'JWT_VERIFY': False,
+    'JWT_SECRET_KEY': SECRET_KEY,
```
Qubad786: Is this Django application secret?

VEDA_OS01/tests/test_transcripts.py
```diff
@@ -1377,3 +1380,184 @@ def test_translations_retrieval_exceptions(self, mock_responses, expected_loggin
...
+                'cielo_username': '',
+                'three_play_secret_key': 'test-secret-key'
+            },
+            None
+        )
+    )
+    @unpack
+    @patch('VEDA_OS01.transcripts.LOGGER')
+    @patch('VEDA_OS01.transcripts.TranscriptCredentialsView.get_cielo_token_response')
+    def test_transcript_credentials(self, post_data, missing_keys, get_cielo_token_mock, mock_logger):
```
Qubad786: we should break it to  `test_transcript_credentials_missing_params` and `test_transcript_credentials_happy_flow`.

VEDA_OS01/tests/test_transcripts.py
```diff
@@ -1377,3 +1380,184 @@ def test_translations_retrieval_exceptions(self, mock_responses, expected_loggin
...
+        else:
+            transcript_credentials = TranscriptCredentials.objects.filter(
+                org=post_data['org'],
+                provider=post_data['provider']
+            )
+            self.assertEqual(len(transcript_credentials), 1)
+
+    @patch('VEDA_OS01.transcripts.LOGGER')
+    @patch('VEDA_OS01.transcripts.TranscriptCredentialsView.get_cielo_token_response')
+    def test_invalid_cielo_credentials(self, get_cielo_token_mock, mock_logger):
```
Qubad786: this would be `test_cielo24_error`.

VEDA_OS01/tests/test_transcripts.py
```diff
@@ -1377,3 +1380,184 @@ def test_translations_retrieval_exceptions(self, mock_responses, expected_loggin
...
+    )
+    @unpack
+    @patch('VEDA_OS01.transcripts.LOGGER')
+    @patch('VEDA_OS01.transcripts.TranscriptCredentialsView.get_cielo_token_response')
+    def test_transcript_credentials(self, post_data, missing_keys, get_cielo_token_mock, mock_logger):
+        """
+        Test that post crednetials works as expected.
+        """
+        # Mock get_cielo_token_mock to return token
+        get_cielo_token_mock.return_value = self.mock_response(
```
Qubad786: We can use `responses` to mock 3rd party requests.

VEDA_OS01/tests/test_transcripts.py
```diff
@@ -1377,3 +1380,184 @@ def test_translations_retrieval_exceptions(self, mock_responses, expected_loggin
...
+    @patch('VEDA_OS01.transcripts.LOGGER')
+    @patch('VEDA_OS01.transcripts.TranscriptCredentialsView.get_cielo_token_response')
+    def test_invalid_cielo_credentials(self, get_cielo_token_mock, mock_logger):
+        """
+        Test that when invalid cielo credentials are supplied, we get correct response.
+        """
+        # Mock get_cielo_token_response.
+        error_type = 'CIELO_LOGIN_FAILURE'
+        error_message = 'Invalid credentials provided to authenticate'
+        get_cielo_token_mock.return_value = self.mock_response(
```
Qubad786: we should use `responses` here too.

VEDA_OS01/transcripts.py
```diff
@@ -360,6 +370,97 @@ def post(self, request, **kwargs):
...
+                '[Transcript Credentials] Required transcript credentials params are missing. %s', ' and '.join(missing)
+            )
+            return Response(
+                status=status.HTTP_400_BAD_REQUEST,
+                data=dict(message=u'{missing} must be specified.'.format(missing=' and '.join(missing)))
+            )
+
+        org = request.data.get('org')
+        api_key = request.data.get('api_key')
+        cielo_username = request.data.get('cielo_username')
```
muzaffaryousaf: This will be simply username.

VEDA_OS01/transcripts.py
muzaffaryousaf: This will be simply secret_key

VEDA_OS01/transcripts.py
```diff
@@ -360,6 +370,97 @@ def post(self, request, **kwargs):
         return Response(status=status.HTTP_200_OK)
 
 
+class TranscriptCredentialsView(APIView):
```
muzaffaryousaf: @nasthagiri Here is the view which will be called from platform.

VEDA/settings.py
```diff
@@ -120,6 +129,7 @@
         'rest_framework.permissions.IsAuthenticated',
     ),
     'DEFAULT_AUTHENTICATION_CLASSES': (
+        'edx_rest_framework_extensions.authentication.JwtAuthentication',
```
muzaffaryousaf: @nasthagiri can you please give it a look, purpose of this PR is to write an endpoint which is useable by platform [PR](https://github.com/edx/edx-platform/pull/16116/commits/5a2e063e24588b0114d4b87cb044f9f7e5c77071#diff-760a096297780408db8ee1ac7b3c9ad4R28)  , We're authenticating via JWT token, just wanted to confirm that this meets edX's standards.

nasthagiri: To minimize unnecessary exposure (per principle of least privilege), I agree that this class should be added to only the views that need JWT auth - so don't add it to the default list.

VEDA/settings.py
nasthagiri: Please don't use django's `SECRET_KEY`.  When that key is used by multiple features, it makes it hard to re-key.  FYI @clintonb - in case he has any best practices to share since he's integrated other edX services using JWT.

VEDA/settings.py
```diff
@@ -120,6 +129,7 @@
...
         'oauth2_provider.ext.rest_framework.OAuth2Authentication',
```
clintonb: Why are you using all of these authentication classes? You only need JWT and session. The others need to go away given that (a) we don't use them anywhere else and (b) they are a potential threat vector.

VEDA/settings.py
```diff
@@ -110,6 +110,15 @@
 # Python dotted path to the WSGI application used by Django's runserver.
```
clintonb: We don't use `settings.py` in any of our IDAs. It probably shouldn't be used here, either. We generally break settings into sub-modules : base, local, devstack, test, production. See https://github.com/edx/cookiecutter-django-ida/tree/master/%7B%7Bcookiecutter.repo_name%7D%7D/%7B%7Bcookiecutter.repo_name%7D%7D/settings.

VEDA/settings.py
```diff
@@ -110,6 +110,15 @@
...
 WSGI_APPLICATION = 'VEDA.wsgi.application'
 
+# See: http://getblimp.github.io/django-rest-framework-jwt/#additional-settings
+JWT_AUTH = {
+    # If the secret is wrong, JWT_VERIFY will raise a jwt.DecodeError telling you as such.
+    # You can still get at the payload by setting the JWT_VERIFY to False
+    'JWT_VERIFY': False,
```
clintonb: Please get rid of this. It should be se to `True` which is the default. If you don't verify the JWT, you are essentially opening the endpoint up to anyone.

VEDA/settings.py
clintonb: See https://github.com/edx/course-discovery/blob/master/course_discovery/settings/base.py#L356-L364.

VEDA/settings.py
clintonb: Have default authentication classes is fine, and is a great way to ensure that your endpoints are always protected. I would keep this here, along with `SessionAuthentication` (useful testing in the browser) The others should go away given that none of them are used.

mushtaqak: @clintonb I am not sure if other authentication classes is used of not. @yro Do you know if there is any authentication classes that need to be removed from here that is not used anymore ?

VEDA/settings.py
mushtaqak: @yro again here, should we get rid of other auth classes? is there any authentication class that should be removed here?

VEDA/settings.py
mushtaqak: You are right @clintonb - We will eventually be dividing the settings into sub parts. This is definitely on TODO list in our future work - CC @muzaffaryousaf

VEDA/settings.py
yro: @mushtaqak The other classes are still in use by the workers.

mushtaqak:
@muzaffaryousaf @clintonb can you guys have look at this again ?

VEDA/settings/base.py
```diff
@@ -110,6 +110,14 @@
 # Python dotted path to the WSGI application used by Django's runserver.
 WSGI_APPLICATION = 'VEDA.wsgi.application'
 
+# NOTE (CCB): JWT_SECRET_KEY is intentionally not set here to avoid production releases with a public value.
```
clintonb: Please don't use my initials for a comment that I did not explicitly write.

requirements.txt
```diff
@@ -13,4 +13,5 @@ pyyaml
 requests==2.18.1
 celery==3.1.18
 pysrt==1.1.1
-edx-opaque-keys==0.4
\ No newline at end of file
+edx-opaque-keys==0.4
+edx-drf-extensions==1.2.3
```
clintonb: Alphabetize these to make future editing/lookup easier.

VEDA/settings/base.py
mushtaqak: Oh, I thought this was some sort of the note that needed to here just like in course-discovery. NVM, I will remove :)

mushtaqak:
@clintonb addressed your feedback.

VEDA/settings/base.py
```diff
@@ -110,6 +110,14 @@
 # Python dotted path to the WSGI application used by Django's runserver.
 WSGI_APPLICATION = 'VEDA.wsgi.application'
 
+# NOTE : JWT_SECRET_KEY is intentionally not set here to avoid production releases with a public value.
```
clintonb: NOTE : —> NOTE: (no space before the colon)

requirements.txt
```diff
@@ -2,9 +2,11 @@
 django==1.9
```
clintonb: Make sure this gets updated to Django 1.11 ASAP. Django 1.9 is EOL.

requirements.txt
```diff
@@ -2,9 +2,11 @@
...
 djangorestframework==3.6.4
 django-cors-headers
-django-oauth-toolkit==0.11.0
-django-model-utils==3.0.0
 django-filter==1.0.4
+django-model-utils==3.0.0
+django-oauth-toolkit==0.11.0
```
clintonb: Why do you need this?

mushtaqak: This was previously added, I just alphabetised as you asked in earlier comment.

clintonb: I understand, but my original question remains: why does this service need an OAuth provider? We already have one on LMS.

mushtaqak: hmm, @yro can tell better I guess

VEDA/settings.py
mushtaqak: My rough guess is that this is being used to communicate with edx-val and video-worker but I am not sure.

requirements.txt
yro: @clintonb There are "dumb" encode worker nodes that call this thing via web API, it's wholly independent from the LMS.

yro: @clintonb To be clear I'm not defending having oauth-toolkit as the "right" solution, just as the "currently in production" solution. I think this may create a little tech debt, but I've got a short term plan to use a different solution for this which will eliminate it.

VEDA/settings.py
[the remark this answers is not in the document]
Qubad786: @mushtaqak why aren't we removing this?

muzaffaryousaf:
@mushtaqak can you rebase this as well ?

mushtaqak:
@muzaffaryousaf Done

instance_config.yaml
```diff
@@ -25,6 +25,14 @@ SECRET_KEY: ""
 # (set to false in prod)
 debug: false
 
+
+# JWT AUTH settings
+JWT_AUTH:
+    JWT_SECRET_KEY':
+    JWT_ISSUER':
+    JWT_AUDIENCE':
+    JWT_VERIFY_AUDIENCE':
```
muzaffaryousaf: @yro  can you please confirm how we'll get the values of these variables on veda production, it should be similar to edx-platform. My main concern is like how / who will get these values from studio production and put them in video pipeline ?

PR #35, opened 2017-09-27
Add video source/speech language for 3PlayMedia.

Qubad786:


Qubad786:
@muzaffaryousaf this is mere implementation, that may change based on conversation going on with 3Play Media.

control/veda_deliver_3play.py
```diff
@@ -94,6 +101,40 @@ def validate_media_url(self):
                 )
             )
 
+    def get_available_languages(self):
+        """
+        Gets available languages
```
muzaffaryousaf: Proper doc string would help to understand & purpose of this method.

control/veda_deliver_3play.py
```diff
@@ -94,6 +101,40 @@ def validate_media_url(self):
...
+        if isinstance(available_languages, dict):
+            raise ThreePlayMediaLanguagesRetrievalError(
+                'Expected 3Play Media Supported languages but got: {response}'.format(response=response.text)
+            )
+
+        return available_languages
+
+    def get_source_language_id(self, languages, source_language_code):
+        """
+        Extracts language id from the given 3Play Media languages
```
muzaffaryousaf: extract which language & why ?

control/veda_deliver_3play.py
```diff
@@ -94,6 +101,40 @@ def validate_media_url(self):
...
             )
 
+    def get_available_languages(self):
+        """
+        Gets available languages
+        """
+        response = requests.get(url=build_url(self.base_url, self.available_languages_url, apikey=self.api_key))
+        if not response.ok:
+            raise ThreePlayMediaLanguagesRetrievalError(
+                'Error while retrieving available languages: {response} -- {status}'.format(
```
muzaffaryousaf: Don't forget to add logger prefix for better tracking.

Qubad786:
@muzaffaryousaf , @muhammad-ammar this is ready for through review.

VEDA_OS01/migrations/0003_auto_20170927_1411.py
```diff
@@ -0,0 +1,25 @@
...
+            model_name='video',
+            name='source_language',
+            field=models.CharField(blank=True, help_text=b'This is only needed with 3Play Media Provider', max_length=50, null=True, verbose_name=b'video source language'),
+        ),
+        migrations.AlterField(
+            model_name='transcriptprocessmetadata',
+            name='lang_code',
+            field=models.CharField(max_length=50, verbose_name=b'Language code'),
+        ),
+    ]
```
muzaffaryousaf: can we have single migration at this point for this feature.

control/veda_deliver_3play.py
```diff
@@ -140,24 +187,27 @@ def generate_transcripts(self):
...
+                '[3PlayMedia] Could not process transcripts for video=%s language=%s.',
                 self.video.studio_id,
+                self.video.source_language,
             )
         except Exception:
             LOGGER.exception(
-                '[3PlayMedia] Unexpected error while transcription for video=%s language=en .',
+                '[3PlayMedia] Unexpected error while transcription for video=%s language=%s.',
                 self.video.studio_id,
+                self.video.source_language,
```
muhammad-ammar: Should `language` be `source_language` in all the above logs so that it is clear from log message that we are talking about  source language?

control/veda_deliver_3play.py
```diff
@@ -111,6 +152,12 @@ def submit_media(self):
             turnaround_level=self.turnaround_level,
             callback_url=self.callback_url,
         )
+
+        available_languages = self.get_available_languages()
+        source_language_id = self.get_source_language_id(available_languages, self.video.source_language)
+        if source_language_id:
+            payload['language_id'] = source_language_id
```
muhammad-ammar: Why we need to do all of above for source language?

control/veda_deliver_3play.py
```diff
@@ -94,6 +101,40 @@ def validate_media_url(self):
                 )
             )
 
+    def get_available_languages(self):
+        """
+        Gets available languages
+        """
+        response = requests.get(url=build_url(self.base_url, self.available_languages_url, apikey=self.api_key))
```
muhammad-ammar: Why we need to do this for every transcription request? Can we cache the response to avoid the request every time?

control/veda_deliver_3play.py
Qubad786: This is just an exception message. Logger prefix is where we are catching it.

control/veda_deliver_3play.py
Qubad786: To avoid errors caused by any changes to 3PlayMedia API.

Qubad786:
@muzaffaryousaf @muhammad-ammar addressed feedbacks & recreated migrations.

PR #41, opened 2017-10-09
Fix drf and django filter versions.

Qubad786:
**Issue**: ImportError: Could not import `'rest_framework.filters.DjangoFilterBackend'`

DRF-3.7.0 has removed `DjangoFilterBackend`, if we want to move to DRF-3.7.0, we shall be using `'django_filters.rest_framework.DjangoFilterBackend'` instead.


Qubad786:
@muzaffaryousaf a quick thumb please?

PR #44, opened 2017-10-12
3Play Media translation process improvements

Qubad786:
[EDUCATOR-1468](https://openedx.atlassian.net/browse/EDUCATOR-1468)
- Translation retrieval process do not use the ranslation endpoint which was previously changed by 3PlayMedia.
- Make a single call to retrieve all the translation metadata for a video.
- Add 100% test coverage for the modified patch.


Qubad786:
@muzaffaryousaf, @muhammad-ammar this is ready.

VEDA_OS01/transcripts.py
```diff
@@ -850,109 +942,112 @@ def retrieve_three_play_translations():
...
+
+                # 5 - if all the processes for this video are complete, update video status in edx-val
+                # update transcript status for video in edx-val as well as edx-video-pipeline.
+                video_jobs = TranscriptProcessMetadata.objects.filter(video=video)
+                if all(video_job.status == TranscriptStatus.READY for video_job in video_jobs):
+                    utils.update_video_status(
+                        val_api_client=val_api,
+                        video=video,
+                        status=TranscriptStatus.READY
+                    )
```
muzaffaryousaf: We can move step 5 in new method.

VEDA_OS01/transcripts.py
```diff
@@ -850,109 +942,112 @@ def retrieve_three_play_translations():
...
-            three_play_secrets.api_key,
-            translation_process.process_id,
-            translation_process.translation_id,
-            translation_process.video.studio_id,
-            translation_process.lang_code,
+        # Retrieve Translations metadata to check the status for each translation.
+        translations = get_translations_metadata(
+            api_key=three_play_secrets.api_key,
+            file_id=file_id,
+            edx_video_id=video.studio_id,
```
muzaffaryousaf: I would extract translation logic into new method. Method with 125 lines won't be acceptable.

Qubad786:
@muzaffaryousaf addressed your feedback.

VEDA_OS01/transcripts.py
```diff
@@ -790,133 +790,200 @@ def three_play_transcription_callback(sender, **kwargs):
...
+def handle_video_translations(video, translations, file_id, api_key, log_prefix):
+    """
+    It is a sub-module of `retrieve_three_play_translations` to handle
+    all the completed translations for a single video.
+
+    Arguments:
+        video: Video data object whose translations need to be handled here.
+        translations: A list containing translations metadata information received from 3play Media.
+        file_id: It is file identifier that is assigned to a Video by 3Play Media.
+        api_key: An api key to communication to the 3Play Media.
```
muhammad-ammar: ` An api key for communicate with the 3Play Media`

Qubad786:
@muzaffaryousaf chaining the `Queryset` will keep building up in a query unless its evaluated, in our case, there will be One DB hit on `.latest()`, see:

screen shot 2017-10-13 at 6 13 43 pm

VEDA_OS01/transcripts.py
```diff
@@ -790,133 +790,200 @@ def three_play_transcription_callback(sender, **kwargs):
...
+        target_language = translation_metadata['target_language_iso_639_1_code']
+
+        if translation_state == COMPLETE:
+            # Fetch the corresponding tracking process.
+            translation_process = get_in_progress_translation_process(
+                video_translation_processes,
+                file_id=file_id,
+                translation_id=translation_id,
+                target_language=target_language
+            )
```
muzaffaryousaf: I would replace this call & call at [line 952](https://github.com/edx/edx-video-pipeline/pull/44/files#diff-f82b0480568faf247efc91479533b6ccR952) with single method call which will return single process, since we don't need all processes.

Qubad786: we need them in case where we need to fail them all.

VEDA_OS01/transcripts.py
```diff
@@ -926,33 +993,90 @@ def retrieve_three_play_translations():
...
+        if not in_progress_translation_processes.exists():
+            LOGGER.info(
+                '[3PlayMedia Task] video=%s does not have any translation process who is in progress.',
+                video.studio_id,
+            )
+            continue
+
+        # Process id remains same across all the processes of a video and its also referred as `file_id`.
+        file_id = in_progress_translation_processes.first().process_id
+        course_id = video.inst_class.local_storedir.split(',')[0]
```
muzaffaryousaf: Since `local_storedir` is optional, can we make sure that this does't break ?

Qubad786:
@muzaffaryousaf addressed.

VEDA_OS01/models.py
```diff
@@ -380,6 +382,23 @@ class Course (models.Model):
...
     )
 
+    @property
+    def org(self):
+        """
+        Get course organization.
+        """
+        org = None
+        course_runs = self.local_storedir
+        if course_runs:
```
muhammad-ammar: I think we don't need to create separate variable for `course_runs`

PR #47, opened 2017-10-16
Debugs from Deployment 16.10.17

yro:
### Description
Debugs from deployment - Oct. 16, 2017

### Testing
- [ ] Unit tests

### Reviewers
- [ ] Code review: @muzaffaryousaf 

### Post-review


muzaffaryousaf:
@yro We've to keep patch build green at-least as it's in master.

muzaffaryousaf:
@yro We've to keep patch build green at-least as it's in master.

muzaffaryousaf:
@yro Are we not using Xuetang now ?

control/veda_deliver.py
```diff
@@ -228,13 +222,22 @@ def _INFORM_INTAKE(self):
...
+        except S3ResponseError:
+            logger.error('[VIDEO_DELIVER] Invalid Storage Bucket')
+            return
+        source_key = bucket.get_key(self.encoded_file)
+        if source_key is None:
+            logger.error('[VIDEO_DELIVER] S3 Intake Object NOT FOUND')
+            return
+
+        source_key.get_contents_to_filename(
+            os.path.join(self.node_work_directory, self.encoded_file)
```
muzaffaryousaf: Why do we need this change, since we're not using the output anywhere ?

muzaffaryousaf:
@yro Are we not using Xuetang now ? If yes, can we also remove veda_deliver_xuetang file and it's settings ?

yro:
@muzaffaryousaf Yes we're done with Xuetang. This is just to bring master to parity with what's on production (I had to do this to get it to run). I'll do a seperate PR to eliminate the xuetang file and do the migrations. 
https://openedx.atlassian.net/browse/EDUCATOR-1499
Will get done before next release.

control/veda_deliver.py
yro: This is the file to reupload to the final endpoint (final s3 bucket, youtube), it has to be downloaded to the worker first. Once we deprecate youtube we can probably refactor this to a simple aws s3 copy command or eliminate the need for this worker to handle files at all, as the encode workers can upload the final encoded files to the endpoint bucket.
This was using wget previously, but the new IAM worker roles are incompatible with it. The stale worker hid this bug previously.

control/veda_deliver.py
```diff
@@ -633,49 +636,6 @@ def start_3play_transcription_process(self):
                 self.video_query.studio_id,
             )
 
-    def _XUETANG_ROUTE(self):
-        if self.video_query.inst_class.xuetang_proc is False:
```
yro: TODO: https://openedx.atlassian.net/browse/EDUCATOR-1499

yro:
@muzaffaryousaf Let me know if this conflicts with your patch build - I'm happy to fix if needed.

control/veda_deliver.py
muzaffaryousaf: We should create ticket for this, this should be done with youtube  deprecate work ?

muzaffaryousaf:
@yro The only concern here is the code coverage dropped. Can you fix it ? It should be like master at-least.

yro:
@muzaffaryousaf Yep, I see it. No problem.

yro:
@muzaffaryousaf I didn't green the patch, but I pushed coverage 5%

muzaffaryousaf:
Can't see coverage jobs for this PR.

yro:
@muzaffaryousaf It was here, then it wasn't -- do you know how to get it back?

control/veda_utils.py
```diff
@@ -275,6 +275,7 @@ def _METADATA(self):
 
     def _FAULT(self, video_object):
         if self.video_object is None:
+            print 'HERE'
```
muzaffaryousaf: This should't be here.

control/veda_val.py
```diff
@@ -91,7 +91,7 @@ def val_tokengen(self):
             'username': self.auth_dict['val_username'],
             'password': self.auth_dict['val_password'],
         }
-
+        print self.auth_dict
```
muzaffaryousaf: No print statement should be used in code. If these are for debug purpose then better to use loggers.

yro: @muzaffaryousaf Sorry - will change, left it in by mistake

control/veda_utils.py
yro: @muzaffaryousaf Sorry - was debugging

control/veda_deliver_xuetang.py
```diff
@@ -1,197 +0,0 @@
...
-   **VEDA Note: since this isn't a real CDN, and represents the
-    'least effort' response to getting video into china,
-    we shan't monitor for success**
-
-
-"""
-auth_dict = get_config()
-
-API_SHARED_SECRET = auth_dict['xuetang_api_shared_secret']
-API_ENDPOINT = auth_dict['xuetang_api_url']
```
muzaffaryousaf: @greg can we also remove this from instance_config ?

control/veda_deliver_xuetang.py
```diff
@@ -1,197 +0,0 @@
...
-
-   **VEDA Note: since this isn't a real CDN, and represents the
-    'least effort' response to getting video into china,
-    we shan't monitor for success**
-
-
-"""
-auth_dict = get_config()
-
-API_SHARED_SECRET = auth_dict['xuetang_api_shared_secret']
```
muzaffaryousaf: @greg can we also remove this from instance_config ?

scripts/youtube_retrieve.py
```diff
@@ -0,0 +1,91 @@
...
+from os.path import expanduser
+import pysftp
+import shutil
+import sys
+
+project_path = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+if project_path not in sys.path:
+    sys.path.append(project_path)
+os.environ['DJANGO_SETTINGS_MODULE'] = 'VEDA.settings'
+django.setup()
```
muzaffaryousaf: @yro  Why do we have to do this in every file, it's not an good thing from Django perspective. Once we understand it's purpose we're planning to remove this from individual files.

scripts/youtube_retrieve.py
```diff
@@ -0,0 +1,91 @@
...
+"""
+import django
+import fnmatch
+import os
+from os.path import expanduser
+import pysftp
+import shutil
+import sys
+
+project_path = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
```
muzaffaryousaf: Why do we have to do this in every file ?

control/veda_deliver.py
```diff
@@ -184,6 +139,29 @@ def run(self):
             if self.video_query.provider == TranscriptProvider.CIELO24:
                 self.cielo24_transcription_flow()
 
+    def hls_run(self):
+        """
+        Get information about encode for URL validation/record
+
+        """
+        self.video_query = Video.objects.filter(edx_id=self.veda_id).latest()
```
muzaffaryousaf: .latest() could break.

control/veda_deliver.py
```diff
@@ -22,34 +28,21 @@
...
-"""
-VEDA Delivery class - determine the destination
-and upload to the appropriate endpoint via the approp. methods
-
-
-"""
 homedir = expanduser("~")
-
-
 watchdog_time = 10.0
```
muzaffaryousaf: Do we need this, given that we've deleted watchdog file

muzaffaryousaf:
@muhammad-ammar  @Qubad786 Can you also take a look at this PR.

control/veda_deliver_xuetang.py
yro: @muzaffaryousaf That't the wrong handle. And yes, I've got it removed in https://github.com/edx/edx-video-pipeline/pull/55

control/veda_deliver_xuetang.py
yro: @muzaffaryousaf It's in a different PR, but I've got it.

scripts/youtube_retrieve.py
yro: @muzaffaryousaf This is just a one-off script, but it needs to find the django project in order to interact with the ORM. This doesn't run as part of the platform, rather, as a sort of management command.

control/veda_deliver.py
yro: How?

control/veda_deliver.py
yro: Ok

scripts/youtube_retrieve.py
yro: @muzaffaryousaf I'll just delete this file entirely.

control/veda_deliver.py
yro: Both latest and get will break potentially, if there's an ID mismatch, but that's ok in this instance, because it's a celery process and it will "fail" and pick up the next task.

yro:
@muzaffaryousaf I'm not going to be able to contribute to this repo for the next few weeks, so deployments are going to be stalled until this is approved.

control/tests/test_deliver.py
```diff
@@ -0,0 +1,169 @@
+"""
+Veda Delivery unit tests
+"""
+from mock import PropertyMock, patch
+import os
+import responses
+import unittest
+
+from django.test import TestCase
```
Qubad786: nit: import order

control/tests/test_deliver.py
```diff
@@ -0,0 +1,169 @@
...
+                val_status='complete'
+            )
+
+    @patch('control.veda_val.VALAPICall._AUTH', PropertyMock(return_value=lambda: CONFIG_DATA))
+    @responses.activate
+    def test_run(self):
+        """
+        Test of HLS run-through function
+        """
+        self.encoded_file = self.veda_id + '_' + self.encode.encode_suffix + '.' + self.encode.encode_filetype
```
Qubad786: use `str.format` instead.

yro:
@muzaffaryousaf This fail is from an outage at edx.org -- it will pass again later

control/tests/test_deliver.py
```diff
@@ -0,0 +1,169 @@
...
+            destination_name='TEST'
+        )
+        self.encode = Encode.objects.create(
+            encode_destination=self.destination,
+            profile_active=True,
+            encode_suffix='HLS',
+            product_spec=self.encode_profile,
+            encode_filetype='HLS'
+        )
+        self.DeliverInstance = VedaDelivery(
```
Qubad786: attributes shouldn't be camel case `self.deliver_instance`.

control/tests/test_deliver.py
```diff
@@ -0,0 +1,169 @@
...
+                self.veda_id + '.m3u8'
+            ))
+        )
+
+    def test_intake(self):
+        """
+        Framework for intake testing
+        """
+        self.DeliverInstance._INFORM_INTAKE()
+        self.encoded_file = self.veda_id + '_' + self.encode.encode_suffix + '.' + self.encode.encode_filetype
```
Qubad786: use `str.format` for string formatting.

control/tests/test_deliver.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def test_validate_url(self):
+        """
+        Test URL Validator
+        """
+        self.assertFalse(self.DeliverInstance._VALIDATE_URL())
+        self.DeliverInstance.endpoint_url = 'https://edx.org'
+        self.assertTrue(self.DeliverInstance._VALIDATE_URL())
+
+    @patch('control.veda_val.VALAPICall._AUTH', PropertyMock(return_value=lambda: CONFIG_DATA))
```
Qubad786: this `patch` is something that should be moved to class level, let's do that.

control/tests/test_deliver.py
```diff
@@ -0,0 +1,169 @@
...
+            veda_id=self.veda_id
+        )
+        with patch.object(VALAPICall, '_AUTH', PropertyMock(return_value=lambda: CONFIG_DATA)):
+            self.VAC = VALAPICall(
+                video_proto=self.VP,
+                val_status='complete'
+            )
+
+    @patch('control.veda_val.VALAPICall._AUTH', PropertyMock(return_value=lambda: CONFIG_DATA))
+    @responses.activate
```
Qubad786: we don't need to activate `responses` for this test case.

control/tests/test_deliver.py
```diff
@@ -0,0 +1,169 @@
...
+            CONFIG_DATA['val_api_url'] + '/XXXXXXXX2014-V00TES1',
+            status=200,
+            json={'error': 'null', 'courses': [], 'encoded_videos': []}
+        )
+        responses.add(responses.PUT, CONFIG_DATA['val_api_url'] + '/XXXXXXXX2014-V00TES1', status=200)
+        self.VP = VideoProto(
+            client_title='Test Video',
+            veda_id=self.veda_id
+        )
+        with patch.object(VALAPICall, '_AUTH', PropertyMock(return_value=lambda: CONFIG_DATA)):
```
Qubad786: Moving such patches to class level won't require this context manager.

control/tests/test_deliver.py
```diff
@@ -0,0 +1,169 @@
...
+            product_spec=self.encode_profile,
+            encode_filetype='HLS'
+        )
+        self.DeliverInstance = VedaDelivery(
+            veda_id=self.veda_id,
+            encode_profile=self.encode_profile,
+            CONFIG_DATA=CONFIG_DATA
+        )
+        # VAL Patching
+        responses.add(responses.POST, CONFIG_DATA['val_token_url'], '{"access_token": "1234567890"}', status=200)
```
Qubad786: These patches should be the part of only those tests that are actually making VAL api calls, not here in `setUp` for every test.

control/tests/test_deliver.py
yro: 👍

control/tests/test_deliver.py
yro: right, sorry - will fix.

control/tests/test_deliver.py
yro: 👍

control/tests/test_deliver.py
yro: @Qubad786 It's failing without it -- it makes the call deeper in the run function, which is a little bit redundant from the `test_update_data` function

control/tests/test_deliver.py
yro: 👍

control/tests/test_deliver.py
yro: 👍

control/tests/test_deliver.py
yro: 👍

control/tests/test_deliver.py
Qubad786: ah, got it, please ignore this one.

control/veda_deliver.py
```diff
@@ -207,13 +184,26 @@ def _INFORM_INTAKE(self):
...
             self.encoded_file
         ))
-        os.system(
-            ' '.join((
-                'wget -O',
-                os.path.join(self.node_work_directory, self.encoded_file),
-                self.hotstore_url
-            ))
+
+        try:
```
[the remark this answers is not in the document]
yro: @Qubad786 I don't disagree, but I wanted to differentiate between the two conditions for config purposes. A better error record for the first condition would probably be `BOTO communication error`

yro:
@Qubad786 This is ready for another review.

control/veda_deliver.py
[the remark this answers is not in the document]
Qubad786: @yro you can log the message that sounds more suitable.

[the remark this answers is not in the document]
yro: 👍

yro:
Thanks @Qubad786

PR #55, opened 2017-10-24
Eliminate XuetangX, Add Migrations

yro:
### Description
Delete XuetangX Code, drop columns.

### Testing
- [ ] Unit tests

### Reviewers
- [ ] Code review: @muzaffaryousaf 

### Post-review


VEDA_OS01/models.py
```diff
@@ -365,15 +365,14 @@ class Course (models.Model):
     )
 
     # TODO: Change field name
-    xue = models.BooleanField('Submit to VAL', default=False)
```
muzaffaryousaf: I won't remove this in the same deployment, for safer side we've to remove the code first and then in next rollout we can remove fields.

yro: @muzaffaryousaf I was doing a "rename", (makemigration prompted for this, so it should be ok) since this field got repurposed. 
It's also, as far as I can tell, not being used anywhere in the codebase. I can definitely hold off if you think that's safer though.

yro: @muzaffaryousaf I should clarify - we have a few places where this is serialized, but it's not being used as a determiner for encodes by the pipeline.

muzaffaryousaf:
@yro That's the way we adopt in other IDA, more you can look into this doc https://openedx.atlassian.net/wiki/spaces/ENG/pages/162953307/Migration+Don+ts

yro:
@muzaffaryousaf Fair enough. I'll save it for another time. Will update.

yro:
@muzaffaryousaf This should go out in its own release, but is blocked by https://github.com/edx/edx-video-pipeline/pull/47 (without that merged to master, the deliver node won't run)

muzaffaryousaf:
@yro can you rebase this PR.

yro:
@muzaffaryousaf Done.

yro:
@muzaffaryousaf @Qubad786 Thanks guys!

Qubad786:
@yro you forgot to rebase before merge, this is have migration conflict now.

yro:
Okay, will fix.

Qubad786:
@yro no need I've fixed in my Course clean PR

PR #59, opened 2017-10-31
update settings override

muhammad-ammar:


VEDA/settings/production.py
```diff
@@ -13,7 +13,7 @@
 
 # Keep track of the names of settings that represent dicts. Instead of overriding the values in base.py,
 # the values read from disk should UPDATE the pre-configured dicts.
-DICT_UPDATE_KEYS = ('DATABASES', 'SECRET_KEY', 'STATIC_ROOT_PATH', 'JWT_AUTH')
```
muhammad-ammar: Note to reviewers: `SECRET_KEY` and `STATIC_ROOT_PATH` are not dicts due to which `vars()[key].update(value)` crashes on line 25 in this file.

mushtaqak:
👍

VEDA/settings/production.py
Qubad786: @mushtaqak now, I see why `DICT_UPDATE_KEYS` were popped from `CONFIG_DATA` 😄

PR #60, opened 2017-11-02
Clean Course Model

Qubad786:
## [EDUCATOR-1667](https://openedx.atlassian.net/browse/EDUCATOR-1667)
    - remove xue, mobile_override, project_ID, s3_dir
    - update course_name, course_hold, institute, class_id fields.


Qubad786:
@muzaffaryousaf @muhammad-ammar take a look.

VEDA_OS01/models.py
```diff
@@ -233,20 +233,15 @@ class Course(models.Model):
...
+    edx_classid = models.CharField('Class ID', max_length=255)
 
     # TODO: Create Default for 'this year' (e.g. 2017)
-    semesterid = models.CharField('Semester', max_length=4)
-    parent_ID = models.CharField(
-        'Parent Project',
-        max_length=8,
-        null=True, blank=True
-    )
+    semesterid = models.CharField('Semester', max_length=4, default='2017')
```
muzaffaryousaf: default should be current year, not the hardcoded one.

VEDA_OS01/models.py
```diff
@@ -233,20 +233,15 @@ class Course(models.Model):
...
-    course_name = models.CharField('Course Name', max_length=100)
+    course_name = models.CharField('Course Name', max_length=255)
 
     # TODO: Change Name (this is reversed)
-    course_hold = models.BooleanField('Course Active', default=False)
-    institution = models.CharField('Inst. Code', max_length=4)
-    edx_classid = models.CharField('Class ID', max_length=5)
+    course_hold = models.BooleanField('Course Active', default=True)
+    institution = models.CharField('Inst. Code', max_length=50)
+    edx_classid = models.CharField('Class ID', max_length=255)
```
muzaffaryousaf: Since, we've visited these fields recently. I'd like to add help text or comment to tell what's the actual purpose/usage of these fields. It'll help others to understand it without any contributor.

Qubad786:
@muzaffaryousaf addressed.

VEDA_OS01/models.py
```diff
@@ -4,13 +4,21 @@
...
 
 from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import CourseKey
 
 
+def current_year():
+    """
+    Returns the current year.
+    """
+    return timezone.now().strftime('%Y')
```
muzaffaryousaf: How about saving the whole datatime instead of year. Also, be aware of .now() VS .now when we're using it as fields in django models.

https://stackoverflow.com/questions/29733203/django-default-datetime-now-in-models-always-saves-same-datetime-after-uwsgi-r

frontend/course_validate.py
```diff
@@ -267,7 +267,6 @@ def course_add(self, return_data):
                     setattr(c1, 'tp_proc', False)
 
         """Always Set"""
-        setattr(c1, 'xue', True)
```
muzaffaryousaf: Do we need to remove these fields from veda_models.yaml as well.

muzaffaryousaf: Also, remove this from destinations.json

control/veda_encode.py
```diff
@@ -70,11 +70,6 @@ def match_profiles(self):
                     self.encode_list.add(e)
                 return None
 
-        if self.course_object.mobile_override is True:
```
muzaffaryousaf: Have a look in veda_models.yaml.

muzaffaryousaf:
@muhammad-ammar I'm done with review, can you also give it a look.

VEDA_OS01/models.py
Qubad786: it is callable.

Qubad786: How about it being Date field instead?

muzaffaryousaf: Yeah that's what i said in first line.

Qubad786:
@muzaffaryousaf have a look, this is tested and ready!

VEDA_OS01/models.py
```diff
@@ -229,23 +237,35 @@ def __unicode__(self):
...
+        help_text='Course number to identify the course.',
+        max_length=255
+    )
+
+    # TODO deprecate semesterid.
+    semesterid = models.CharField(
+        'Semester',
+        max_length=4,
+        help_text='The course creation year.',
+        default=current_year,
```
muzaffaryousaf: Leave it as it, if we're planning to deprecate this field.

Qubad786:
Going to merge.

VEDA_OS01/models.py
```diff
@@ -229,24 +229,31 @@ def __unicode__(self):
         )
 
 
-class Course(models.Model):
+class Course(TimeStampedModel):
     """
     Model for Course.
```
muzaffaryousaf: @Qubad786 I meant that we should add the comments about this course i.e. it's not a course run but a whole course.

Qubad786: @muzaffaryousaf added.

Qubad786:
@yro have a look at 2nd commit.

PR #61, opened 2017-11-06
heartbeat endpoint

muhammad-ammar:
[EDUCATOR-1632](https://openedx.atlassian.net/browse/EDUCATOR-1632)

### DIFF
--- a/VEDA_OS01/tests/test_views.py
+++ b/VEDA_OS01/tests/test_views.py
@@ -1,10 +1,11 @@
 """ Views tests """
 import json
-import responses
 
+import responses
 from ddt import data, ddt, unpack
 from django.contrib.auth.models import User
 from django.core.urlresolvers import reverse
+from django.db.utils import DatabaseError
 from mock import patch
 from rest_framework import status
 from rest_framework.test import APITestCase
@@ -221,3 +222,26 @@ class TranscriptCredentialsTest(APITestCase):
             json.dumps({'error': error_message}),
             status.HTTP_400_BAD_REQUEST
         )
+
+
+class HeartbeatTests(APITestCase):
+    """
+    Tests for hearbeat endpoint.
+    """
+    def test_heartbeat(self):
+        """
+        Test that heartbeat endpoint gives expected response upon success.
+        """
+        response = self.client.get(reverse('heartbeat'))
+        assert response.status_code == 200
+        assert json.loads(response.content) == {'OK': True}
+
+    @patch('django.db.backends.utils.CursorWrapper')
+    def test_heartbeat_failure_db(self, mocked_cursor_wrapper):
+        """
+        Test that heartbeat endpoint gives expected response when there is an error.
+        """
+        mocked_cursor_wrapper.side_effect = DatabaseError
+        response = self.client.get(reverse('heartbeat'))
+        assert response.status_code == 500
+        assert json.loads(response.content) == {'OK': False}
--- a/VEDA_OS01/views.py
+++ b/VEDA_OS01/views.py
@@ -2,25 +2,27 @@
 
 import json
 import logging
+
 import requests
-
-from django.http import HttpResponse
-from django.http import HttpResponseRedirect
+from django.db import connection
+from django.db.utils import DatabaseError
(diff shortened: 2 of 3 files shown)


muhammad-ammar:
@muzaffaryousaf @mushtaqak @Qubad786 This is ready for review.

VEDA_OS01/views.py
```diff
@@ -284,3 +286,36 @@ def user_login(request):
...
+    """
+    View to check if database is reachable and ready to handle requests.
+    """
+    try:
+        db_status()
+    except DatabaseError as fail:
+        return JsonResponse(
+            {
+                'OK': False,
+                'check': 'db',
```
Qubad786: reminder: only keep `{'OK': False}`

VEDA_OS01/tests/test_views.py
```diff
@@ -221,3 +222,27 @@ def test_cielo24_error(self, mock_logger):
...
+    @patch('django.db.backends.utils.CursorWrapper')
+    def test_heartbeat_failure_db(self, mocked_cursor_wrapper):
+        """
+        Test that heartbeat endpoint gives expected response when there is an error.
+        """
+        error = 'Unable to connect to db'
+        mocked_cursor_wrapper.side_effect = DatabaseError(error)
+        response = self.client.get(reverse('heartbeat'))
+        assert response.status_code == 500
+        assert json.loads(response.content) == {'OK': False, 'check': 'db', 'error': error}
```
Qubad786: we can use `self.assertDict` here.

VEDA_OS01/tests/test_views.py
```diff
@@ -221,3 +222,27 @@ def test_cielo24_error(self, mock_logger):
...
+    """
+    Tests for hearbeat endpoint.
+    """
+    def test_heartbeat(self):
+        """
+        Test that heartbeat endpoint gives expected response upon success.
+        """
+        response = self.client.get(reverse('heartbeat'))
+        assert response.status_code == 200
+        assert json.loads(response.content) == {'OK': True}
```
Qubad786: we can use `self.assertDict` here.

VEDA_OS01/views.py
```diff
@@ -284,3 +286,36 @@ def user_login(request):
...
+
+@api_view(['GET'])
+@permission_classes([AllowAny])
+def heartbeat(request):  # pylint: disable=unused-argument
+    """
+    View to check if database is reachable and ready to handle requests.
+    """
+    try:
+        db_status()
+    except DatabaseError as fail:
```
mushtaqak: I would prefer `error` or `exception`

VEDA_OS01/views.py
mushtaqak: Why are we sending `check: db` ?

VEDA_OS01/tests/test_views.py
```diff
@@ -221,3 +222,27 @@ def test_cielo24_error(self, mock_logger):
...
+
+    @patch('django.db.backends.utils.CursorWrapper')
+    def test_heartbeat_failure_db(self, mocked_cursor_wrapper):
+        """
+        Test that heartbeat endpoint gives expected response when there is an error.
+        """
+        error = 'Unable to connect to db'
+        mocked_cursor_wrapper.side_effect = DatabaseError(error)
+        response = self.client.get(reverse('heartbeat'))
+        assert response.status_code == 500
```
mushtaqak: Nice :)

VEDA_OS01/tests/test_views.py
muhammad-ammar: For pytests, we should use assert statement. @mushtaqak your thoughts, you were discussing something like this yesterday?

muhammad-ammar:
@mushtaqak @Qubad786 Addressed your feedback

VEDA_OS01/tests/test_views.py
Qubad786: @muhammad-ammar If we are already inheriting from `APITestCase` anyway, why shouldn't we use `self.assertXXX..`?

PR #66, opened 2017-11-15
separate static configuration

muhammad-ammar:
This will move the static configuration into its own yaml file. We need this to avoid updating ansible role every time we introduce a new variable. From now onwards please add all static configuration in `static_config.yaml`

### DIFF
--- a/VEDA/utils.py
+++ b/VEDA/utils.py
@@ -8,8 +8,9 @@ import yaml
 from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import CourseKey
 
-DEFAULT_CONFIG_FILE_PATH = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+CONFIG_ROOT_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 DEFAULT_CONFIG_FILE_NAME = 'instance_config.yaml'
+STATIC_CONFIG_FILE_PATH = os.path.join(CONFIG_ROOT_DIR, 'static_config.yaml')
 
 
 def extract_course_org(course_id):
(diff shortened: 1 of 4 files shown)


muhammad-ammar:
@muzaffaryousaf @mushtaqak @yro @Qubad786 FYI

instance_config.yaml
```diff
@@ -136,66 +121,3 @@ sg_script_key:
...
-    hls:
-        - hls
-
-#--
-# Heal settings
-heal_start: 1
-heal_end: 144
-global_timeout: 40
-
-...
```
mushtaqak: Three dots mystery 😆

Qubad786: finally resolved.

VEDA/utils.py
```diff
@@ -61,14 +62,17 @@ def get_config(yaml_config_file=DEFAULT_CONFIG_FILE_NAME):
...
             yaml_config_file
         )
 
     with open(yaml_config_file, 'r') as config:
         config_dict = yaml.load(config)
 
-    return config_dict
+    with open(STATIC_CONFIG_FILE_PATH, 'r') as config:
+        static_config_dict = yaml.load(config)
+
```
mushtaqak: I think, there should be a comment here.

VEDA/utils.py
```diff
@@ -8,8 +8,9 @@
 from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import CourseKey
 
-DEFAULT_CONFIG_FILE_PATH = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+CONFIG_FILE_PATH = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
```
Qubad786: CONFIG_ROOT_DIR would be better? As currently, its giving an impression as if it is config file path but its config dir path.

PR #67, opened 2017-11-15
Refine HEAL process, make logging louder

yro:
### Description
Change heal queue, make logging for deliver/heal louder

### Testing
- [ ] Unit tests

### Reviewers
- [ ] Code review: Qubad786

### Post-review

### DIFF
--- a/bin/deliver
+++ b/bin/deliver
@@ -61,7 +61,7 @@ class DeliverCli:
                     'worker',
                     '--loglevel=info',
                     '--concurrency=' + str(auth_dict['celery_threads']),
-                    '-Q ' + auth_dict['celery_deliver_queue'],
+                    '-Q ' + auth_dict['celery_deliver_queue'] + ',' + auth_dict['celery_heal_queue'],
                     '-n deliver.%h'
                 ))
             )
--- a/instance_config.yaml
+++ b/instance_config.yaml
@@ -105,7 +105,7 @@ val_username:
 val_transcript_create_url:
 val_video_transcript_status_url:
 
-
+# Celery Worker Config Information
 rabbitmq_broker:
 rabbitmq_pass:
 rabbitmq_user:
--- a/bin/heal
+++ b/bin/heal
@@ -34,7 +34,7 @@ class HealCli:
                       .replace(hour=0, minute=0, second=0, microsecond=0) \
                       .astimezone(pytz.utc) + timedelta(days=1)
 
-        maintainer_healer.apply_async((self.binscript,), queue='transcode_stat', eta=go_time)
+        maintainer_healer.apply_async((self.binscript,), queue=auth_dict['celery_heal_queue'], eta=go_time)
 
 
 def main():
--- a/control/veda_deliver.py
+++ b/control/veda_deliver.py
@@ -8,6 +8,7 @@ import datetime
 import logging
 import shutil
 from os.path import expanduser
+import sys
 
 import boto
 import boto.s3
@@ -35,6 +36,9 @@ except ImportError:
     from veda_deliver_3play import ThreePlayMediaClient
 
 LOGGER = logging.getLogger(__name__)
+# TODO: Remove this temporary logging to stdout
+logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+
 try:
     boto.config.add_section('Boto')
 except:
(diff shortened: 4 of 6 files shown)


bin/deliver
```diff
@@ -61,7 +61,7 @@ class DeliverCli:
                     'worker',
                     '--loglevel=info',
                     '--concurrency=' + str(auth_dict['celery_threads']),
-                    '-Q ' + auth_dict['celery_deliver_queue'],
+                    '-Q ' + auth_dict['celery_deliver_queue'] + ',' + 'heal_queue',
```
Qubad786: @yro should this be in `auth_dict` ??

instance_config.yaml
```diff
@@ -105,7 +105,17 @@ val_username:
...
+# Celery Info
+# ---
+celery_app_name:
+# can do multiple queues like so: foo,bar,baz
+celery_worker_queue: encode_worker,large_encode_worker
+celery_deliver_queue: deliver_worker
+celery_heal_queue: heal_queue
+celery_threads: 1
+
+onsite_worker: False
```
muhammad-ammar: @yro We have separated configuration into two yaml files. Please see https://github.com/edx/edx-video-pipeline/pull/66. You need to move/update these values in [static_config.yaml](https://github.com/edx/edx-video-pipeline/blob/master/static_config.yaml). Let us know if there are any issues.

yro: @muhammad-ammar OK! Some of these aren't static, but that's a smart move. I'll make the change.

yro:
@Qubad786 @muhammad-ammar This is ready for another review.

instance_config.yaml
```diff
@@ -105,7 +105,9 @@ val_username:
 val_transcript_create_url:
 val_video_transcript_status_url:
 
-
+# Celery Worker Config Information
+onsite_worker: False
+celery_threads: 1
```
muhammad-ammar: @yro Why we still need these two new variables here? It would be great if we can move these into `static_config.yaml`. I am saying this because the problem is that once the ansible roles will start working this `instance_config.yaml `file be written by ansible, so introduction of any new variable in this file means that we also need to change the ansible role which is very impractical.

Qubad786: @yro how are these settings are configurable? i.e. What will be their values on sandbox/production environment? Will their values remain same no matter the environment (sandbox/prod)? if so, we should make them static.

Qubad786:
@yro lets try to get done with this and start the deployment right away, as we've to test secure credentials work on production before we share "go through" plan with partners.

instance_config.yaml
Qubad786: Moreover, if these settings are configurable, then let @muhammad-ammar know their environment-specific values.

yro:
@Qubad786 I cannot start deployment until this is merged.

instance_config.yaml
yro: @Qubad786 @muhammad-ammar That's the wrong approach. If this file cannot be configured individually for workers without having to change several repos, this codebase becomes almost unusable. This config file should be copied from the veda-secure repo in whole.

Qubad786: @yro veda-secure should only contain secure configs (e.g. secret_keys, access_keys etc), not all the configurations are meant to be secure.

yro: @Qubad786 Why make multiple things when there can be one? It seems like having multiple files that need to be changed per instance and use case make the service less operable.

Qubad786: So, `static_config.yaml`  should contain non-secure + static configurations. I hope this clears up our perspective.

yro: So, let's say I have a larger worker instance I want to deploy. Right now I change a single file. You're saying that I should have to change two files.

muhammad-ammar: @yro You can merge the code the way best suitable as we want to release the work as soon as possible. 

But we need to resolve this. I have explained the reason in my first comment above. Can you please spare some time and explain why we need to change/copy this file or any other file every time. It would be great if you can share the details, then I can discuss here with my team and after we can have a discussion with you to improve this whole flow. 
What do you say?

Qubad786: @yro the answer to your question lies in @muhammad-ammar 's first comment on the top of the conversation.

yro: @muhammad-ammar I'll move these to static, so as to unblock you -- they can be changed later if we want.

muhammad-ammar: @yro I am not blocked on anything. I divided the config into two for ansible work. But that work will take time to finish. So you can do thing they way easy for you now. As I said in my last comment we will discuss about this.

yro: 👍 Sounds good! We will discuss.

yro:
@muhammad-ammar @Qubad786 👍 Thanks guys.

PR #71, opened 2017-11-23
improve url construction

muhammad-ammar:
https://openedx.atlassian.net/browse/EDUCATOR-1822


muhammad-ammar:
@mushtaqak @Qubad786 This is ready for review.

Qubad786:
@muhammad-ammar does this also include fix to `scrub_url`?

control/tests/test_deliver_cielo.py
```diff
@@ -146,13 +146,13 @@ def test_transcript_flow(self):
                     v=CIELO24_API_VERSION,
                     job_id='000-111-222',
                     target_language='TARGET_LANG',
-                    callback_url=build_url(
+                    callback_url='{}?job_id={}&iwp_name={}&lang_code={}&org={}&video_id={}'.format(
```
mushtaqak: I think we should use format variables here like `{cielo_callback_base_url}?job_id={job_id}&iwp_name={iwp_name}&lang_code={lang_code}&org={org}&video_id={video_id}`

control/veda_deliver_cielo.py
```diff
@@ -117,13 +117,13 @@ def perform_transcript(self, job_id, lang_code):
         """
         Request cielo24 to generate transcripts for a video.
         """
-        callback_url = build_url(
+        callback_url = '{}?job_id={}&iwp_name={}&lang_code={}&org={}&video_id={}'.format(
```
mushtaqak: same as above

control/veda_deliver_cielo.py
```diff
@@ -136,7 +136,7 @@ def perform_transcript(self, job_id, lang_code):
             api_token=self.api_key,
             priority=self.turnaround,
             transcription_fidelity=self.fidelity,
-            options=json.dumps({"return_iwp": ["FINAL"]})
+            options=json.dumps({"return_iwp":["FINAL"]})
```
mushtaqak: wasn't space better for quality ?

control/tests/test_deliver_cielo.py
muhammad-ammar: This will make the line longer and pylint will complain.

control/veda_deliver_cielo.py
muhammad-ammar: This will make the line longer and pylint will complain.

control/veda_deliver_cielo.py
muhammad-ammar: it was causing issues during url encoding so we removed it

PR #75, opened 2017-12-07
add service variant name

muhammad-ammar:


VEDA/settings/production.py
```diff
@@ -7,16 +7,18 @@
 
 DEBUG = False
 TEMPLATE_DEBUG = DEBUG
+DEFATULT_SERVICE_VARIANT_NAME = 'video-pipeline'
```
mushtaqak: Don't we need this in case of local environment or test environment?

muhammad-ammar: I don't think so.

PR #79, opened 2017-12-19
Improve logging and add an option to run 3play translations.

Qubad786:
[EDUCATOR-1979](https://openedx.atlassian.net/browse/EDUCATOR-1979)

Improve logging and add an option to run 3play translations with heal.

### DIFF
--- a/bin/heal
+++ b/bin/heal
@@ -51,7 +51,8 @@ def main():
         {cmd} -i veda_id
         {cmd} -c course_id
         {cmd} -s schedule
-        [-i -c -s]
+        {cmd} --process-translations
+        [-i -c -s --process-translations]
         Use --help to see all options.
         '''.format(cmd=sys.argv[0])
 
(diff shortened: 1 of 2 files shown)


Qubad786:
@mushtaqak @muzaffaryousaf quick review?

bin/heal
```diff
@@ -71,12 +72,27 @@ def main():
...
 
     veda_id = args.veda_id
     course_id = args.course_id
     schedule = args.schedule
 
+    process_translations = args.process_translations
+    if process_translations:
+        # Only kick off a round of retrieving successful
+        # translations from 3Play Media
+        os.environ['DJANGO_SETTINGS_MODULE'] = 'VEDA.settings.production'
```
mushtaqak: Do we need to add this here ?

Qubad786: yes. it is to tell Django about the correct settings.

Qubad786: This is temporary and will be removed once we have Django mgmt. command for this.

Qubad786:
Going to merge.

PR #80, opened 2018-01-03
Start transcription once video encodings(except for "review" and "youtube") get ready

muhammad-ammar:
[EDUCATOR-2033](https://openedx.atlassian.net/browse/EDUCATOR-2033)
Start transcription when a video encodings are ready (except for "review" and "youtube" encodes).

### DIFF
--- a/VEDA_OS01/tests/factories.py
+++ b/VEDA_OS01/tests/factories.py
@@ -1,0 +1,88 @@
+"""
+VEDA model factories.
+"""
+from factory import Sequence, SubFactory
+from factory.django import DjangoModelFactory
+
+from VEDA_OS01.models import Course, Destination, Encode, TranscriptStatus, URL, Video, VideoStatus
+
+
+class CourseFactory(DjangoModelFactory):
+    """
+    Course data model factory.
+    """
+    class Meta(object):
+        model = Course
+
+    course_name = Sequence('Test Course {0}'.format)
+    institution = Sequence('INST-{0}'.format)
+    edx_classid = Sequence('CLASS-{0}'.format)
+    semesterid = Sequence('2018-{0}'.format)
+    proc_loc = False
+    review_proc = False
+    yt_proc = False
+    s3_proc = False
+    local_storedir = None
+
+
+class VideoFactory(DjangoModelFactory):
+    """
+    Video data model factory.
+    """
+    class Meta(object):
+        model = Video
+
+    inst_class = SubFactory(CourseFactory)
+    client_title = Sequence('Video {0}'.format)
+    edx_id = Sequence('ABC-CDE-EFG-{0}'.format)
+    studio_id = Sequence('61bd0526{0}'.format)
+    video_trans_status = VideoStatus.SI
+    transcript_status = TranscriptStatus.NOT_APPLICABLE
+    process_transcription = False
+    provider = None
+    three_play_turnaround = None
+    cielo24_turnaround = None
+    cielo24_fidelity = None
+    source_language = None
+    preferred_languages = []
+
+
+class DestinationFactory(DjangoModelFactory):
+    """
+    Destination data model factory.
+    """
+    class Meta(object):
+        model = Destination
+
+    destination_name = Sequence('Dest-{0}'.format)
+    destination_active = False
+    destination_nick = Sequence('D{0}'.format)
+
+
+class EncodeFactory(DjangoModelFactory):
+    """
+    Encode data model factory.
+    """
+    class Meta(object):
+        model = Encode
+
+    encode_destination = SubFactory(DestinationFactory)
+    encode_name = Sequence('Encode-{0}'.format)
+    profile_active = False
+    encode_suffix = ''
+    encode_filetype = 'mp4'
+    encode_bitdepth = None
+    encode_resolution = None
+    product_spec = None
+
+
+class UrlFactory(DjangoModelFactory):
+    """
+    URL data model factory.
+    """
+    class Meta(object):
+        model = URL
+
+    encode_profile = SubFactory(EncodeFactory)
+    videoID = SubFactory(VideoFactory)
+    encode_url = Sequence('https://www.querty.com/{0}'.format)
--- a/VEDA_OS01/utils.py
+++ b/VEDA_OS01/utils.py
@@ -1,8 +1,8 @@
 """
 Common utils.
 """
-
-from VEDA_OS01.models import TranscriptStatus
+from VEDA.utils import get_config
+from VEDA_OS01.models import Encode, TranscriptStatus, URL, Video
 
 
 class ValTranscriptStatus(object):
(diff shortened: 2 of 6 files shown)


control/veda_deliver.py
```diff
@@ -101,7 +101,9 @@ def run(self):
         """
         if self.encode_profile == 'youtube':
             self._CLEANUP()
+            self.start_transcription(self._DETERMINE_STATUS())
             return None
+
         if self.encode_profile == 'review':
             return None
```
Qubad786: what about this encoding?

control/veda_deliver.py
```diff
@@ -127,17 +129,34 @@ def run(self):
...
+        """
+        Kick off the transcription process
+
+        Arguments:
+            status (str): `Complete` or `Progress`
+        """
+        # We only want to generate transcripts when all encodings are completed
+        if status == 'Complete' and self.video_query.process_transcription:
+            encode_query = Encode.objects.get(
+                product_spec='desktop_mp4'
```
Qubad786: @yro is there any course who does not have mp4 encoding active?

control/veda_deliver.py
```diff
@@ -127,17 +129,34 @@ def run(self):
...
+    def start_transcription(self, status):
+        """
+        Kick off the transcription process
+
+        Arguments:
+            status (str): `Complete` or `Progress`
+        """
+        # We only want to generate transcripts when all encodings are completed
+        if status == 'Complete' and self.video_query.process_transcription:
+            encode_query = Encode.objects.get(
```
Qubad786: `encode_profile` would be better?

muhammad-ammar:
@yro Can you also look at these changes? What we are trying to do is that start transcription process only when all video encodings are completed. Do you have any concerns?

control/veda_deliver.py
muhammad-ammar: @yro can you share some detail about this?

yro:
@muhammad-ammar I think it's okay to start the transcription process before all the encodings are completed. Youtube especially is subject to delays, and will artificially increase the wait time for the transcript, so firing it before everything is done is probably best

control/veda_deliver.py
yro: Everything gets an mp4 encoding, so leave this one off. We're trying to deprecate using youtube altogether, and a video would have already triggered transcripting via its mp4 encode, so this would be a duplicate.

muhammad-ammar:
> I think it's okay to start the transcription process before all the encodings are completed. Youtube especially is subject to delays, and will artificially increase the wait time for the transcript, so firing it before everything is done is probably best

Actually the problem is that statuses in val get overridden so thats why we are starting transcription after all videos completed their encodings. Please see [EDUCATOR-2033](https://openedx.atlassian.net/browse/EDUCATOR-2033). 

Here is a scenario, some video encodings finished and some are in progress, we started transcription process and set the `Transcription in Progress` status on val, after that, in progress video finished encoding and it also updated the `Ready` status on val. Now this will be confusing because a course author expect `Transcript Ready` after `Transcription in Progress` but the status is `Ready`

muhammad-ammar:
> Get rid of the youtube encode call (we don't need it), and this is great!

Can you please mention which call you are referring to?

muhammad-ammar:
@yro Please read my replies to your comments above ^^

control/veda_deliver.py
yro: @muhammad-ammar This is the call I'm referring to -- since everything gets an mp4 encoding, this would be a duplicate call (right? I could be mistaken)

yro:
@muhammad-ammar Youtube is always last to complete, if the course has youtube turned on. The 'complete' status is changed by the youtube callback function. My concern is that this wouldn't get called at all if the course has youtube turned on. Is it possible to override the other status checks to ignore and not change the status if there's a transcript status?

control/veda_deliver.py
Qubad786: @yro if you see the PR diff, we are not kicking transcription process in case of mp4 encode anymore. 

A little summary, we don't want any video encoding override Transcription statuses in Studio once the transcription has started. To achieve this, we have to wait for encodings to complete before we start transcription (Because if we do transcription on mp4 encoding then any other encoding that comes after mp4 encoding will overwrite video status from "transcript in progress" to something else which is undesirable and spread confusions among authors).

Do you think of any other approach that we could adopt instead?

Qubad786:
@yro we cannot make status checks to not to overwrite status if there is a transcript status, this is because if youtube encoding has failed and status is some transcript status then that 'yt encoding failed' does not show up studio (which is again confusing).

Qubad786:
@muhammad-ammar @mushtaqak this is ready for review.

VEDA_OS01/utils.py
```diff
@@ -57,3 +57,53 @@ def invalidate_fernet_cached_properties(model, fields):
...
+    course = video.inst_class
+    # Pick the encodes map from the settings.
+    encodes_map = get_config().get('encode_dict', {})
+    # Active encodes according to course instance.
+    for attr, encodes in encodes_map.iteritems():
+        if getattr(course, attr, False):
+            encode_list += [encode.strip() for encode in encodes]
+
+    # Filter active encodes further according to their corresponding encode profiles activation.
+    for encode in list(encode_list):
```
muhammad-ammar: `encode_list` is already list then why `list(encode_list)`?

VEDA_OS01/utils.py
```diff
@@ -57,3 +57,53 @@ def invalidate_fernet_cached_properties(model, fields):
...
+        encode_profile = Encode.objects.filter(product_spec=encode).first()
+        if not encode_profile or (encode_profile and not encode_profile.profile_active):
+            encode_list.remove(encode)
+
+    # Filter encodes based on their successful encoding for the specified video.
+    for encode in list(encode_list):
```
muhammad-ammar: same comment as above

VEDA_OS01/utils.py
```diff
@@ -57,3 +57,53 @@ def invalidate_fernet_cached_properties(model, fields):
...
+    return encode_list
+
+
+def is_video_ready(edx_id, exceptions=list()):
+    """
+    Check whether a video should be considered ready.
+
+    Arguments:
+        edx_id: an ID identifying the VEDA video.
+        exceptions: A list containing the profiles that should not be considered.
```
muhammad-ammar: arguments types and `Returns` part is missing?

VEDA_OS01/utils.py
```diff
@@ -57,3 +57,53 @@ def invalidate_fernet_cached_properties(model, fields):
...
+            videoID=video,
+            encode_profile__product_spec=encode
+        )
+        if completed_encode_profile.exists():
+            encode_list.remove(encode)
+
+    return encode_list
+
+
+def is_video_ready(edx_id, exceptions=list()):
```
muhammad-ammar: can we change `exceptions` with `ignored_encodes` or `exclude_encodes` or something better? `exceptions` is confusing and at first it seems that it will contain the python exception types.

VEDA_OS01/utils.py
Qubad786: to make a copy of it.

Qubad786:
feedback addressed.

VEDA_OS01/utils.py
```diff
@@ -57,3 +57,53 @@ def invalidate_fernet_cached_properties(model, fields):
...
+            videoID=video,
+            encode_profile__product_spec=encode
+        )
+        if completed_encode_profile.exists():
+            encode_list.remove(encode)
+
+    return encode_list
+
+
+def is_video_ready(edx_id, ignore_encodes=list()):
```
mushtaqak: `ignore_encodes=[]` ?

VEDA_OS01/utils.py
```diff
@@ -57,3 +57,53 @@ def invalidate_fernet_cached_properties(model, fields):
...
+    """
+    Check whether a video should be considered ready.
+
+    Arguments:
+        edx_id(unicode): an ID identifying the VEDA video.
```
mushtaqak: typo: `An ID ...`

control/veda_deliver.py
```diff
@@ -127,17 +133,34 @@ def run(self):
         self._UPDATE_DATA()
         self._CLEANUP()
 
-        # Transcription Process
-        # We only want to generate transcripts for `desktop_mp4` profile.
-        if self.encode_profile == 'desktop_mp4' and self.video_query.process_transcription:
+        # We only want to generate transcripts when all the encodings(except for YT and Review) are done.
+        if utils.is_video_ready(self.video_query.edx_id, ignore_encodes=['review', 'youtube']):
```
mushtaqak: I think we should have this check inside `start_transcription` method.

VEDA_OS01/utils.py
Qubad786: Nope. This would be considered as pass by reference.

VEDA_OS01/tests/factories.py
```diff
@@ -0,0 +1,88 @@
...
+class UrlFactory(DjangoModelFactory):
+    """
+    URL data model factory.
+    """
+    class Meta(object):
+        model = URL
+
+    encode_profile = factory.SubFactory(EncodeFactory)
+    videoID = factory.SubFactory(VideoFactory)
+    encode_url = factory.sequence('https://www.querty.com/{0}'.format)
```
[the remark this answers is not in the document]
Qubad786: nice catch I wonder that why did tests not fail.

VEDA_OS01/tests/factories.py
```diff
@@ -0,0 +1,88 @@
+"""
+VEDA model factories.
+"""
+import factory
```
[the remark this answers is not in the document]
Qubad786: hmm ok.

control/veda_deliver.py
Qubad786: I was doing that before, it kinda feel dirty.

VEDA_OS01/tests/factories.py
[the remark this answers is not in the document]
Qubad786: just found out the reason, `factory.sequence` is a decorator.

Qubad786:
@muhammad-ammar @mushtaqak feedbacks addressed.

muhammad-ammar:
👍

mushtaqak:
👍

PR #85, opened 2018-02-05
Yro/EDUCATOR-2152

yro:
### Description
Attempt to clean up and streamline enqueueing for celery workers, as well as collapse encode-finding to a single method.

### Testing
- [x] Unit tests

### Reviewers
- [ ] Code review: @muhammad-ammar @Qubad786 

FYI: @edx/educator-neem

### Post-review
- [x] Rebase and squash commits

### DIFF
--- a/control/tests/test_heal.py
+++ b/control/tests/test_heal.py
@@ -107,49 +107,43 @@ class HealTests(TestCase):
 
     @data(
         {
-            'edx_id': '1',
             'video_trans_status': 'Corrupt File',
             'video_trans_start': datetime.datetime.utcnow().replace(tzinfo=utc),
             'video_active': True,
+            'expected_encodes': set(['hls', 'mobile_low'])
         },
         {
-            'edx_id': '1',
             'video_trans_status': 'Review Reject',
             'video_trans_start': datetime.datetime.utcnow().replace(tzinfo=utc),
             'video_active': True,
+            'expected_encodes': []
         },
         {
-            'edx_id': '1',
             'video_trans_status': 'Review Hold',
             'video_trans_start': datetime.datetime.utcnow().replace(tzinfo=utc),
             'video_active': True,
+            'expected_encodes': []
         },
         {
-            'edx_id': '2',
             'video_trans_status': 'Complete',
             'video_trans_start': datetime.datetime.utcnow().replace(tzinfo=utc),
             'video_active': False,
+            'expected_encodes': set(['hls', 'mobile_low'])
         },
         {
-            'edx_id': '2',
             'video_trans_status': 'Ingest',
             'video_trans_start': datetime.datetime.utcnow().replace(tzinfo=utc),
             'video_active': True,
-        },
-        {
-            'edx_id': '1',
-            'video_trans_status': 'Corrupt File',
-            'video_trans_start': datetime.datetime.utcnow().replace(tzinfo=utc),
-            'video_active': True,
+            'expected_encodes': set(['hls', 'mobile_low'])
         },
     )
     @unpack
-    def test_determine_fault(self, edx_id, video_trans_status, video_trans_start, video_active):
+    def test_determine_fault(self, video_trans_status, video_trans_start, video_active, expected_encodes):
         """
         Tests that determine_fault works in various video states.
         """
         video_instance = Video(
-            edx_id=edx_id,
+            edx_id='test_id',
             video_trans_status=video_trans_status,
             video_trans_start=video_trans_start,
             video_active=video_active,
@@ -157,13 +151,69 @@ class HealTests(TestCase):
         )
         video_instance.save()
 
-        encode_list = self.heal_instance.determine_fault(video_instance)
+        encodes = self.heal_instance.determine_fault(video_instance)
+        self.assertEqual(encodes, expected_encodes)
 
-        if video_instance.edx_id == '1':
-            self.assertEqual(encode_list, [])
-        elif video_instance.edx_id == '2':
-            for e in encode_list:
-                self.assertTrue(e in self.encode_list)
+    @data(
+        {
+            'video_trans_status': 'Corrupt File',
+            'video_trans_start': datetime.datetime.utcnow().replace(tzinfo=utc),
+            'video_active': True,
+            'expected_encodes': []
+        },
+        {
+            'video_trans_status': 'Review Reject',
+            'video_trans_start': datetime.datetime.utcnow().replace(tzinfo=utc),
+            'video_active': True,
+            'expected_encodes': []
+
+        },
+        {
+            'video_trans_status': 'Review Hold',
+            'video_trans_start': datetime.datetime.utcnow().replace(tzinfo=utc),
+            'video_active': True,
+            'expected_encodes': []
+
+        },
+        {
+            'video_trans_status': 'Complete',
+            'video_trans_start': datetime.datetime.utcnow().replace(tzinfo=utc),
+            'video_active': False,
+            'expected_encodes': set(['hls', 'mobile_low'])
+
+        },
+        {
+            'video_trans_status': 'Ingest',
+            'video_trans_start': datetime.datetime.utcnow().replace(tzinfo=utc),
+            'video_active': True,
+            'expected_encodes': set(['hls', 'mobile_low'])
+
+        },
+        {
+            'video_trans_status': 'Corrupt File',
+            'video_trans_start': datetime.datetime.utcnow().replace(tzinfo=utc),
+            'video_active': True,
+            'expected_encodes': []
+
+        },
+    )
+    @unpack
+    def test_determine_fault_freeze_bug(
+            self, video_trans_status, video_trans_start, video_active, expected_encodes
+    ):
+        video_instance = Video(
+            edx_id='test_id',
+            video_trans_status=video_trans_status,
+            video_trans_start=video_trans_start,
+            video_active=video_active,
+            inst_class=self.course_object
+        )
+        video_instance.save()
+
+        heal_instance_two = VedaHeal(freezing_bug=True)
+        encode_list = heal_instance_two.determine_fault(video_instance)
+        self.assertEqual(encode_list, expected_encodes)
+
 
     @data(
         {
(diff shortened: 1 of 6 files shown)


static_config.yaml
```diff
@@ -5,14 +5,6 @@
...
 edx_s3_rejected_prefix: prod-edx/rejected/
 
-# Celery Info
-onsite_worker: False
-celery_threads: 1
-celery_app_name: veda_production
-celery_worker_queue: encode_worker,large_encode_worker
-celery_deliver_queue: deliver_worker
-celery_heal_queue: heal_queue
-
```
muhammad-ammar: why we have deleted all these? this will cause crashes for new sandboxes because those sandboxes expect these settings in this file.

test_config.yaml
```diff
@@ -24,6 +24,16 @@ three_play_api_transcript_url: https://static.3playmedia.com/
...
 
+# ---
+# Celery Info
+# ---
+onsite_worker: False
+celery_threads: 1
+celery_app_name: veda_test_app
+celery_worker_queue: test_encode_worker
+celery_deliver_queue: test_deliver_worker
+celery_heal_queue: test_heal_queue
```
muhammad-ammar: we don't need these settings in this file.

control/veda_file_ingest.py
```diff
@@ -243,10 +184,8 @@ def database_record(self):
             self.full_filename += "." + self.video_proto.file_extension
 
         if not os.path.exists(self.full_filename):
-            ErrorObject().print_error(
-                message='Ingest: File Not Found'
-            )
-            return None
+            LOGGER.exception('[VIDEO_INGEST] File Not Found {veda_id}', veda_id=self.video_proto.veda_id)
```
muhammad-ammar: I think this will not work. this should be 

```python
LOGGER.exception('[VIDEO_INGEST] File Not Found %s', self.video_proto.veda_id)
```

control/veda_file_ingest.py
```diff
@@ -385,12 +324,12 @@ def abvid_report(self):
...
             return None
 
-        R = Report(
+        email_rept = Report(
             status="File Corrupt on Ingest",
             upload_serial=self.video_proto.abvid_serial,
             youtube_id=''
         )
-        R.upload_status()
+        email_rept.upload_status()
```
muhammad-ammar: `email_report` would be much better.

control/veda_heal.py
```diff
@@ -85,7 +84,7 @@ def send_encodes(self):
                     jobid = uuid.uuid1().hex[0:10]
                     celeryapp.worker_task_fire.apply_async(
                         (veda_id, encode_profile, jobid),
-                        queue=self.auth_dict['celery_worker_queue'].split(',')[0]
+                        queue=self.auth_dict['celery_worker_queue']
```
muhammad-ammar: this always confuses me. `celery_worker_queue` is defined as comma separated string with 2 queue names in it. how the multiple queue string will work in the above case? why we don't want to use `.split(',')[0]`? Shouldn't we have only single queue name for `celery_worker_queue` variable?

control/veda_file_ingest.py
```diff
@@ -130,55 +106,22 @@ def insert(self):
...
-            )
-
-        """
-        Update Video Status
-        """
-        Video.objects.filter(
-            edx_id=self.video_proto.veda_id
-        ).update(
-            video_trans_status='Queue'
+        encode_instance.send_encodes()
```
muhammad-ammar: do we need to return some value from `send_encodes` and then act accordingly here? the below log message will always be logged whatever happens in `send_encodes` method, is this correct?

control/veda_file_ingest.py
```diff
@@ -130,55 +106,22 @@ def insert(self):
...
-                queue=self.auth_dict['celery_worker_queue'].split(',')[0]
-            )
-
-        """
-        Update Video Status
-        """
-        Video.objects.filter(
-            edx_id=self.video_proto.veda_id
-        ).update(
-            video_trans_status='Queue'
```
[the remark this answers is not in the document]
yro: @muhammad-ammar The method called updates it.

control/celeryapp.py
```diff
@@ -28,7 +28,7 @@
     rabbitmq_broker=auth_dict['rabbitmq_broker']
 )
 
-app = Celery(auth_dict['celery_app_name'], broker=CEL_BROKER, backend=CEL_BACKEND, include=[])
+app = Celery(auth_dict.get('celery_app_name', None), broker=CEL_BROKER, backend=CEL_BACKEND, include=[])
```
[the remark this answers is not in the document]
yro: @muhammad-ammar I was trying to move the celery information out of static-config since it's fungible, but I see your point about sandboxes below, so I'll put it back.

control/veda_file_ingest.py
yro: Good point. There's logging in the submethod. Deleting.

control/veda_file_ingest.py
yro: Right, sorry, should be .format or string formatting. Fixed.

control/veda_heal.py
yro: @muhammad-ammar We should only have one queue name. This was a holdover from a previous iteration where we had different queues for filesizes, but we don't need it any more (we just upped the EC2 instance storage.)
It's weird, because celery consumers and celery enqueuers read the string differently, and it's cause enqueueing problems in the past. 
So, one queue. That should make the machines be less confused.

control/veda_heal.py
```diff
@@ -98,8 +97,9 @@ def determine_fault(self, video_object):
                 self.val_status = 'file_corrupt'
                 return []
 
-        if video_object.video_trans_status == 'Review Reject' or video_object.video_trans_status == 'Review Hold' or \
-            video_object.video_trans_status == 'Review Hold':
+        if video_object.video_trans_status == 'Review Reject' or \
+                video_object.video_trans_status == 'Review Hold' or \
+                video_object.video_trans_status == 'Review Hold':
```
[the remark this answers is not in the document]
yro: 👍

static_config.yaml
yro: Yep! I was trying to move these settings, but point taken. Replacing.

test_config.yaml
yro: @muhammad-ammar Yep, I see above.

yro:
@muhammad-ammar This is ready for review.

control/veda_heal.py
Qubad786: @muhammad-ammar https://github.com/edx-ops/veda-secure/pull/11

Qubad786: @yro `celery_worker_queue` is unchanged in the `static_config.yaml` and static configs overwrite the `intstance_config.yaml` today. So, this will crash is production.

control/tests/test_heal.py
```diff
@@ -165,6 +165,62 @@ def test_determine_fault(self, edx_id, video_trans_status, video_trans_start, vi
...
+            edx_id=edx_id,
+            video_trans_status=video_trans_status,
+            video_trans_start=video_trans_start,
+            video_active=video_active,
+            inst_class=self.course_object
+        )
+        video_instance.save()
+        heal_instance_two = VedaHeal(freezing_bug=True)
+        encode_list = heal_instance_two.determine_fault(video_instance)
+        if video_instance.edx_id == '1':
```
Qubad786: @yro we should avoid branch on `ddt` while asserting the expectations. 

As long as we pass the expectations along with the `ddt`, we wouldn't be falling in a situation where we need to branch on `ddt` for asserting the expectations. 

May you please address this. Think about just having `self.assertItemsEqual(actual_encode_list, ddt_passed_expectations)` instead of `the if/else`s.

muhammad-ammar: @yro Please see the below [test](https://github.com/edx/edx-video-pipeline/blob/master/VEDA/tests/test_utils.py#L71-L110) to see how we should write ddt in unit tests. Let us know if you need any help. 

```python
    @data(
        {
            'urls': ('http://api.cielo24/', '/add/job'),
            'params': {},
            'expected_url': 'http://api.cielo24/add/job'
        },
        {
            'urls': ('http://api.cielo24', '/add/job'),
            'params': {'a': 1, 'b': 2},
            'expected_url': 'http://api.cielo24/add/job?a=1&b=2'
        },
        {
            'urls': ('http://api.cielo24/', 'add/job'),
            'params': {'c': 3, 'd': 4},
            'expected_url': 'http://api.cielo24/add/job?c=3&d=4'
        },
        {
            'urls': ('http://api.cielo24', 'add/job'),
            'params': {'p': 100},
            'expected_url': 'http://api.cielo24/add/job?p=100'
        },
        {
            'urls': ('http://api.cielo24', 'add/job', 'media'),
            'params': {'p': 100},
            'expected_url': 'http://api.cielo24/add/job/media?p=100'
        }
    )
    @unpack
    def test_build_url(self, urls, params, expected_url):
        """
        Tests that utils.build_url works as expected.
        """
        url = utils.build_url(
            *urls,
            **params
        )
        self.assertEqual(
            url,
            expected_url
        )
```

control/veda_heal.py
yro: @Qubad786  This is exactly the situation I was trying to avoid. We shouldn't have this be a static config. What if we want to have different priority queues in the future? This is a flaw, but I'll make the change to get this shipped.

control/tests/test_heal.py
yro: @Qubad786 OK.

yro:
@Qubad786 @muhammad-ammar Feedback addressed.

control/veda_heal.py
yro: Right on -- I'll leave it for now. Thanks!

control/tests/test_heal.py
```diff
@@ -107,63 +107,113 @@ def test_heal(self):
...
+            edx_id='test_id',
+            video_trans_status=video_trans_status,
+            video_trans_start=video_trans_start,
+            video_active=video_active,
+            inst_class=self.course_object
+        )
+        video_instance.save()
+
+        encodes = self.heal_instance.determine_fault(video_instance)
+        self.assertEquals(encodes, expected_encodes)
```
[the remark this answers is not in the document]
Qubad786: `assertEquals` is a deprecated alias, please use `assertEqual` instead.
Ref: https://docs.python.org/2/library/unittest.html#deprecated-aliases

PR #90, opened 2018-02-28
Restore old transcription workflow

Qubad786:
[EDUCATOR-2408](https://openedx.atlassian.net/browse/EDUCATOR-2408)

This restores old transcription flow when credentials were stored with VEDA course.

### DIFF
--- a/control/old_veda_deliver_cielo.py
+++ b/control/old_veda_deliver_cielo.py
@@ -1,0 +1,173 @@
+import requests
+import ast
+import urllib
+
+"""
+Cielo24 API Job Start and Download
+Options (reflected in Course.models):
+transcription_fidelity =
+  Mechanical (75%),
+  Premium (95%)(3-72h),
+  Professional (99+%)(3-72h)
+priority =
+  standard (24h),
+  priority (48h)
+turnaround_hours = number, overrides 'priority' call, will change a standard to a priority silently
+"""
+from control_env import *
+from veda_utils import ErrorObject
+
+requests.packages.urllib3.disable_warnings()
+
+
+class Cielo24TranscriptOld(object):
+
+    def __init__(self, veda_id):
+        self.veda_id = veda_id
+        '''Defaults'''
+        self.c24_site = 'https://api.cielo24.com/api'
+        self.c24_login = '/account/login'
+        self.c24_joblist = '/job/list'
+        self.c24_newjob = '/job/new'
+        self.add_media = '/job/add_media'
+        self.transcribe = '/job/perform_transcription'
+
+        '''Retreive C24 Course-based defaults'''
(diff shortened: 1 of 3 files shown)


Qubad786:
@muhammad-ammar @mushtaqak @irfanuddinahmad may you guys give it a look?

control/veda_deliver.py
```diff
@@ -629,6 +635,72 @@ def start_3play_transcription_process(self, encoded_file):
                 self.video_query.studio_id,
             )
 
+    def _CIELO24_UPLOAD(self):
+        """
+        Note: This is part of the old flow which was deprecated when transcript phase 1 was reased.
+        """
+        # TODO: this must be removed once existing 3rd party credentials are migrated according
```
mushtaqak: @Qubad786 Would there be any issue now we are reverting this change?

control/veda_deliver.py
```diff
@@ -128,6 +129,11 @@ def run(self):
         u1.encode_size = self.video_proto.filesize
         u1.save()
 
+        # TODO: Warning! this shall be removed once 3rd party credentials
```
mushtaqak: I think, we should create a follow-up ticket that tracks this and when 3rd party Automated Video Transcription phases are rolled out, we would need to clean this again.

@irfanuddinahmad FYI

control/veda_deliver.py
```diff
@@ -5,10 +5,10 @@
 
 """
 import datetime
+import ftplib
 import logging
 import shutil
 from os.path import expanduser
-import sys
```
mushtaqak: We are using `sys` in the code below.

control/old_veda_deliver_cielo.py
```diff
@@ -0,0 +1,173 @@
...
+        print str(r4.status_code) + ' : Cielo24 Status Code'
+        return ast.literal_eval(r4.text)['TaskId']
+
+
+def main():
+    pass
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
mushtaqak: `sys` is not imported

control/veda_deliver.py
Qubad786: we have it: https://openedx.atlassian.net/browse/EDUCATOR-1321

control/old_veda_deliver_cielo.py
Qubad786: its coming from `from control_env import *`

control/veda_deliver.py
Qubad786: its coming from `from control_env import *`

control/veda_deliver.py
Qubad786: nope, but we have to make the credentials are migrated.

PR #109, opened 2018-06-19
Add 'no_audio' optional flag to HEAL command

ssemenova:
Added 'no_audio' optional flag to HEAL command, to skip over mp3 encoding if a video has no audio.


schenedx:
Deploy this change to the delivery node and see if this helps us?

PR #110, opened 2018-06-26
EmailAlert should not send email in staging environment

ssemenova:
Re-wrote EmailAlert so it doesn't send an email in the staging environment, also cleaned up the code to be a little less redundant.

Part of this series of PRs to create a VEDA staging environment:
[Terraform](https://github.com/edx/terraform/pull/831)
[veda-secure](https://github.com/edx-ops/veda-secure/pull/19)


control/veda_file_discovery.py
```diff
@@ -32,6 +32,7 @@
 
 logging.basicConfig(level=logging.INFO)
 logging.getLogger("requests").setLevel(logging.WARNING)
+logging.getLogger("boto").setLevel(logging.CRITICAL)
```
schenedx: I think it's enough to do `ERROR` level instead?

ssemenova: You're probably right. Changing it

PR #114, opened 2018-07-26
Remove polling ingest

ssemenova:
Remove polling ingest, since we have HTTP ingest now.


bin/loop.py
```diff
@@ -66,27 +66,9 @@ def run(self):
         """
         actually run the function
         """
-        if self.ingest is True:
```
schenedx: Please also remove the parser.add_argument lines associated with ingest on line 50

PR #115, opened 2018-07-26
Fixing ingest race condition

ssemenova:
Fixing ingest race condition, where two videos get saved under the same video number.

### DIFF
--- a/control/veda_file_ingest.py
+++ b/control/veda_file_ingest.py
@@ -5,6 +5,7 @@ Discovered file ingest/insert/job triggering
 import datetime
 import logging
 import subprocess
+import threading
 
 from django.db.utils import DatabaseError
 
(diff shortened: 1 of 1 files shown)


control/veda_file_ingest.py
```diff
@@ -216,17 +218,18 @@ def database_record(self):
         * Note: defensive against the possibility of later passing in an ID
         """
         if self.video_proto.veda_id is None:
-            lsid = self.course_object.last_vid_number + 100
-            self.video_proto.veda_id = self.course_object.institution
-            self.video_proto.veda_id += self.course_object.edx_classid
-            self.video_proto.veda_id += self.course_object.semesterid
-            self.video_proto.veda_id += "-V" + str(lsid).zfill(6)
+            with threading.RLock():
+                lsid = self._get_last_vid_number() + 100
```
schenedx: Put this into a function please and name it clearly.

PR #122, opened 2018-08-16
VEDA Heal webpage

ghs180:


templates/heal.html
```diff
@@ -0,0 +1,20 @@
...
+    ------V E D A------
+    <form method="post" action="">
+      {% csrf_token %}
+      <br />
+      Video ID: <input type="text" name="veda_id"><br>
+      <input type="submit" value="Heal">
+    </form>
+
+    {% if result %}
+      <p>Current status: {{ result }}</p>
```
schenedx: `Current Status` is the wrong message in my opinion.
You could say: `Heal run result:`

instance_config.yaml
```diff
@@ -1,115 +0,0 @@
----
```
schenedx: Why is this file removed? Any reason for doing so?

PR #125, opened 2018-09-06
Fix about page ingest name problem

schenedx:
@yro Please review.
With this change, we can handle about video names like 
`Fisica II _ Vincenzo Canale & Francesco Bloisi _ Trailer.mp4`


yro:
👍

PR #127, opened 2018-09-14
Fix the about video ingest with names containing brackets

schenedx:
@yro Please review
We had another about video ingest breakage, because the video name to be ingested was
`MOOC_Promo_004_Valoración_Financiera_Comprimido (1).m4v`. Note the bracket in the name. That introduced problems with ffprobe

### DIFF
--- a/control/veda_file_ingest.py
+++ b/control/veda_file_ingest.py
@@ -131,7 +131,9 @@ class VedaIngest(object):
             .replace('\'', '\\\'') \
             .replace('[', '\[') \
             .replace(']', '\]') \
-            .replace('&', '\&')
+            .replace('&', '\&') \
+            .replace('(', '\(') \
+            .replace(')', '\)')
 
         ff_command = ' '.join((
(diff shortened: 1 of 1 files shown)


control/veda_file_ingest.py
```diff
@@ -131,7 +131,9 @@ def _gather_metadata(self):
             .replace('\'', '\\\'') \
             .replace('[', '\[') \
             .replace(']', '\]') \
-            .replace('&', '\&')
+            .replace('&', '\&') \
+            .replace('(', '\(') \
+            .replace(')', '\)')
```
yro: @schenedx It might make sense to make a catch-all function for these that replace them with mystery-characters. 
You could decode the chars iteratively and then just replace them with `?` or something. This is fine though.

schenedx: I know what you are saying. This is a slippery slope. 
I am not prepared to sink time into these right now. But yes, this is one of the area in the code we should improve.

PR #129, opened 2018-10-15
ARCH-255: Update edx-drf-extensions imports.

robrap:
ARCH-255

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,3 +1,4 @@
 Mushtaq Ali <mushtaak@gmail.com>
 Muhammad Ammar <mammar@gmail.com>
 Gregory Martin <greg@edx.org>
+Robert Raposa <rraposa@edx.org>
--- a/VEDA/settings/base.py
+++ b/VEDA/settings/base.py
@@ -119,7 +119,7 @@ DATABASES = {
 JWT_AUTH = {
     'JWT_AUDIENCE': 'video-pipeline',
     'JWT_ISSUER': 'video-pipeline',
-    'JWT_DECODE_HANDLER': 'edx_rest_framework_extensions.utils.jwt_decode_handler',
+    'JWT_DECODE_HANDLER': 'edx_rest_framework_extensions.auth.jwt.decoder.jwt_decode_handler',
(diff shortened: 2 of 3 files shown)


nasthagiri:
@schenedx Looks like your team has recently pushed changes to this repo.  The Arch team would like to upgrade one of its dependent libraries.  Questions:

1. The master branch (and hence this PR as well) is failing on Travis.  The underlying issue doesn't seem to have been solved for several months.  Should we ignore those errors and move on?

nasthagiri:
FYI @Qubad786 - see comment above - in case you're still working in this area.

schenedx:
@nasthagiri Yes. We should ignore the errors and merge.
That said. this also mean we should talk about auth change in veda. Even if we merge this change, it doesn't mean this change will be automatically deployed to production. The deployment is pure manual. I need to know a date for this change to be deployed.

nasthagiri:
@robrap Given @schenedx's response, we can go ahead and merge this PR - assuming other related dependencies are also added if needed.

robrap:
@schenedx: All tests from the following file are failing:
```
control/tests/test_file_discovery.py FFFFFFFFFF
```
I can merge my change, but I'd like to propose that the following be done:
Add a ticket titled "Video Pipeline has no test coverage for file discovery".  That can point to the PR where test_file_discovery.py gets deleted because they are all failing.  This will at least get master back to Green so the rest of your tests can protect the repo.  Thoughts?

schenedx:
@robrap That's not the only test file which contains failed tests. That said, I am fine having a ticket to fix all the unit tests associated with this repo. Created at EDUCATOR-3570

robrap:
Removed failing tests in separate PR (https://github.com/edx/edx-video-pipeline/pull/130). :)
FYI: @nasthagiri

robrap:
@schenedx: FYI: I'll leave the codecov settings as-is.  That way PR will continue to be red due to the very low codecov, but at least the Tracis CI builds will be green.  This particular change doesn't require a special deployment.  As @nasthagiri mentioned, we will need to plan a deployment for some future work.

PR #150, opened 2018-12-28
Add priority queues and re-route normal flow to medium queue

Qubad786:


instance_config.yaml
```diff
@@ -96,7 +96,9 @@ veda_token_url:
 # ---
 celery_app_name:
 # can do multiple queues like so: foo,bar,baz
-celery_worker_queue:
+celery_worker_high_queue:
```
schenedx: Can you comment on this line to say this queue is not being used and will be used when we wire up the functionality from studio?

Qubad786: sure.

PR #155, opened 2019-05-29
Azarembok/veda heal

azarembok:
This PR enables the veda heal tool (including very simple UI accessible from the /heal URL) and support within the delivery worker.  Changes here include those made by other developers that had not been merged, plus some minor fixes from me.

### DIFF
--- a/control/veda_heal.py
+++ b/control/veda_heal.py
@@ -19,7 +19,6 @@ from django.utils.timezone import utc
 from VEDA_OS01.models import Encode, URL, Video
 from VEDA_OS01.utils import VAL_TRANSCRIPT_STATUS_MAP
 
-import celeryapp
 from control_env import WORK_DIRECTORY, HEAL_START, HEAL_END
 from veda_encode import VedaEncode
 from veda_val import VALAPICall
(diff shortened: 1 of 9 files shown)


azarembok:
@fysheets please hold off on reviewing for now, I will need some time to work out the build issues.

azarembok:
@fysheets the build/test issues look sorted out, you can review this when you get a chance.  I'll squash my commits before merging.

control/veda_heal.py
```diff
@@ -106,18 +105,20 @@ def send_encodes(self):
                 veda_id = v.edx_id
                 encode_profile = encode
                 job_id = uuid.uuid1().hex[0:10]
-                task_result = celeryapp.worker_task_fire.apply_async(
-                    (veda_id, encode_profile, job_id),
-                    queue=self.auth_dict['celery_worker_medium_queue'].strip(),
-                    connect_timeout=3
-                )
+                # TODO: Commenting this out temporarily.  The worker_task_fire task
```
fysheets: For the official record, why did this need to be commented out / what was breaking with it included?

templates/heal.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
```
fysheets: This shouldn't be too hard to make prettier :)

control/veda_heal.py
azarembok: We had a circular dependency between veda_heal.py and celeryapp.py (both were importing each other), and this was causing errors when a 3rd module tried to import veda_heal.  To remove the dependency on celeryapp I removed the call above.

PR #162, opened 2019-06-25
PROD-411: Move ingest processing to celery task.  

azarembok:
PROD-411: Move ingest and upload to hotstore of course videos to celery task.  Prevents api call from AWS from timing out and retrying, which results in extra encode requests.

### DIFF
--- a/control/http_ingest_celeryapp.py
+++ b/control/http_ingest_celeryapp.py
@@ -1,0 +1,64 @@
+"""
+Start Celery Worker
+"""
+
+from __future__ import absolute_import
+
+from celery import Celery
+import logging
+import sys
+import boto
+from boto.exception import S3ResponseError
+
+from VEDA.utils import get_config
+from control.veda_file_discovery import FileDiscovery
+
+
+LOGGER = logging.getLogger(__name__)
+# TODO: Remove this temporary logging to stdout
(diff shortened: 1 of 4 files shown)


control/http_ingest_celeryapp.py
```diff
@@ -0,0 +1,64 @@
...
+@app.task(name=auth_dict['celery_http_ingest_queue'])
+def ingest_video_and_upload_to_hotstore(requested_key):
+    LOGGER.info('ingest_video_and_upload_to_hotstore key %s' % requested_key)
+    bucket_name = auth_dict['edx_s3_ingest_bucket']
+    try:
+        connection = boto.connect_s3()
+        bucket = connection.get_bucket(bucket_name)
+        s3_key = bucket.get_key(requested_key)
+    except S3ResponseError:
+        LOGGER.error('[INGEST CELERY TASK] Could not connect to S3')
```
fysheets: Can we please add the `requested_key` to this logging file so that way if we're looking for a failed key in splunk this error message will be returned.

worker.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash -x
+#--- Startup Script for VEDA HTTP Ingest Worker --#
```
fysheets: Can we rename this file to something a bit more descriptive so that we know what it's a script to do

PR #171, opened 2019-07-31
PROD-509:  Fix 500 error after submitting heal request.

azarembok:
Also fixes PROD-414:  Changed text on VEDA heal page to indicate that VEDA id is required.

### DIFF
--- a/templates/heal.html
+++ b/templates/heal.html
@@ -17,8 +17,8 @@
         <form method="post" action="">
           {% csrf_token %}
           <div class="form-group">
-            <label for="veda_id">Video ID</label>
-            <input type="text" name="veda_id" class="form-control" id="veda_id" aria-describedby="veda_id" placeholder="Enter Video ID">
+            <label for="veda_id">VEDA Video ID</label>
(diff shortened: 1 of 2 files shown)


templates/heal.html
```diff
@@ -17,8 +17,8 @@ <h1 align="center">VEDA Heal Tool</h1>
         <form method="post" action="">
           {% csrf_token %}
           <div class="form-group">
-            <label for="veda_id">Video ID</label>
-            <input type="text" name="veda_id" class="form-control" id="veda_id" aria-describedby="veda_id" placeholder="Enter Video ID">
+            <label for="veda_id">VEDA Video ID</label>
```
fysheets: Does this change close out the ticket in our backlog to clarify what ID to include in this tool?

azarembok: Yes it does...I mentioned it in the first PR comment above.

fysheets: :) Just noticed it, when I clicked into the PR from Github it took me straight to the code and not to the home page. Thanks!

PR #175, opened 2019-09-23
Added code to find videos whose source is not present in our hotstore.

azarembok:
This was used for data analysis after completing the HLS re-encode process,
saving here for posterity.


azarembok:
Merging without review since this code was already run in production and served its intended purpose.  It is only run as part of a management command that we may never use again.

PR #176, opened 2019-09-23
PROD-549: Use Django Oauth Toolkit for authentication.

azarembok:
Changes the authentication method used within edx-video-pipeline to use Django Oauth Toolkit (DOT) for authentication, due to the previous method (DOP) being deprecated.

### DIFF
--- a/control/tests/test_val.py
+++ b/control/tests/test_val.py
@@ -1,12 +1,11 @@
 
 from __future__ import absolute_import
-import ast
 import os
 import sys
 from django.test import TestCase
 from ddt import data, ddt, unpack
 
-from mock import PropertyMock, patch
+from mock import Mock, PropertyMock, patch
 
 import requests
 import urllib3
@@ -31,7 +30,6 @@ CONFIG_DATA = utils.get_config('test_config.yaml')
 
 @ddt
 class TestVALAPI(TestCase):
-
     def setUp(self):
         self.VP = VideoProto(
             client_title='Test Title',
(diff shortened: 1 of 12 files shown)


control/veda_val.py
```diff
@@ -9,7 +9,10 @@
 import urllib3
 import ast
 import json
+import datetime
 
+from django.core.cache import cache
+from edx_rest_api_client.client import EdxRestApiClient
```
doctoryes: `EdxRestApiClient` is actually deprecated.
See https://github.com/edx/edx-video-worker/pull/55#discussion_r328193241

azarembok: @doctoryes I have updated to use OAuthAPIClient, please review when you get a chance.  Thanks!

control/tests/test_val.py
```diff
@@ -6,11 +6,14 @@
...
 
-from mock import PropertyMock, patch
+from mock import Mock, PropertyMock, patch
 
 import requests
 import urllib3
 import responses
+import datetime
+
+from edx_rest_api_client.client import EdxRestApiClient
```
doctoryes: This line can be removed.

control/veda_val.py
```diff
@@ -202,14 +181,7 @@ def send_val_data(self):
...
 
-        r1 = requests.get(
-            '/'.join((
-                self.auth_dict['val_api_url'],
-                self.video_proto.val_id
-            )),
-            headers=self.headers,
-            timeout=self.auth_dict['global_timeout']
-        )
+        r1 = self.oauth2_client.request('GET', '/'.join((self.auth_dict['val_api_url'], self.video_proto.val_id, '')))
```
doctoryes: The trailing `/` didn't exist before - is this OK?

control/veda_val.py
```diff
@@ -332,13 +304,7 @@ def send_404(self):
...
         )
 
-        r2 = requests.post(
-            self.auth_dict['val_api_url'] + '/',
-            data=json.dumps(sending_data),
-            headers=self.headers,
-            timeout=self.auth_dict['global_timeout']
-        )
-
+        r2 = self.oauth2_client.request('POST', self.auth_dict['val_api_url'] + '/', data=json.dumps(sending_data))
```
doctoryes: Nit: It'd be nice to keep a standard way of forming the URL string in all these calls - all using `join`.

control/veda_val.py
```diff
@@ -58,54 +60,33 @@ def __init__(self, video_proto, val_status, **kwargs):
...
-
-        if r.status_code != 200:
-            LOGGER.error('[API] : VAL Token generation')
-            return
-
-        self.val_token = ast.literal_eval(r.text)['access_token']
-        self.headers = {
-            'Authorization': 'Bearer ' + self.val_token,
-            'content-type': 'application/json'
-        }
```
doctoryes: 👍 for this code simply melting away...

control/veda_val.py
azarembok: Thanks for catching this...I don't think it hurts but it doesn't need to be there.  I will remove.

control/veda_val.py
azarembok: 👍

PR #184, opened 2019-10-15
Fix string/byte handling for python3 compatibility.

azarembok:


azarembok:
> Looks good to me. I'm assuming status_bar was removed cause it wasn't used anywhere??

Correct...it was unused so I deleted it rather than try to fix it.  Thanks for the review!

PR #215, opened 2020-05-07
Added support for python 3.8 and dropped support for older Django versions

mraarif:
Relevant JIRA issue [here](https://openedx.atlassian.net/browse/BOM-1603)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: python
 
 python:
-  - "3.7"
+  - 3.7
+  - 3.8
 
 env:
   - TOXENV=django22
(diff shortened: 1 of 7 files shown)


mraarif:
@jmbowman should we make all these changes just to add python 3.5(I've had to downgrade some packages and add these decode statements for 3.5 specifically)?, previously this repo was using 3.7 only and I'm assuming it should be fine upgrading it to 3.8 directly, what do you think?

jmbowman:
If it's already using 3.7 in production, then correct - no reason to add 3.5 support, just add 3.8 and keep testing 3.7.

PR #218, opened 2020-06-02
move JWT header prefix settings under JWT_AUTH

DawoudSheraz:
### [PROD-1507](https://openedx.atlassian.net/browse/PROD-1507)

JWT_AUTH_HEADER_PREFIX is JWT_AUTH settings and should be defined under that dict


DawoudSheraz:
@azarembok Will the same change be needed in veda-secure instance config? I am not sure if the veda-secure YAML file overrides or updates JWT_AUTH dict

